PR #65, opened 2018-02-22
Add test for github

nightfury1204:


lib/github_test.go
```diff
@@ -0,0 +1,491 @@
...
+		}
+	}
+	return nil
+}
+
+func verifyAuthenticatedTokenReview(review *v1.TokenReview, teamSize int) error {
+	if !review.Status.Authenticated {
+		return fmt.Errorf("Expected authenticated ture, got false")
+	}
+	if review.Status.User.Username != "nahid" {
```
tamalsaha: Define constants for these values

lib/github_test.go
```diff
@@ -0,0 +1,491 @@
...
+}
+`
+	)
+
+	teamSizes := []int{25, 0, 1, 13, 100, 77, 233}
+
+	for _, size := range teamSizes {
+		// page=1, PerPage=25
+		// authenticated : true
+		t.Run("scenario 1", func(t *testing.T) {
```
tamalsaha: Change the test name to `"team size: %d"`

lib/github_test.go
```diff
@@ -0,0 +1,491 @@
...
+   "user":{
+      "login":"nahid",
+      "id":1204
+   }
+}
+`
+	)
+
+	teamSizes := []int{25, 0, 1, 13, 100, 77, 233}
+
```
tamalsaha: Restructure the test data like https://github.com/kubernetes/kubernetes/pull/54366/files#diff-11ca6bae99dd41cf90ffc5528959802dR175 , so that the actual test code becomes simple

PR #66, opened 2018-02-26
Add test for gitlab

nightfury1204:


lib/gitlab_test.go
```diff
@@ -3,47 +3,364 @@ package lib
...
+	}
+	err := g.Client.SetBaseURL(serverUrl)
 	if err != nil {
-		return nil, errors.Errorf("failed to read `%s`.Reason: %v", name, err)
+		return nil, err
+	}
+	return g, nil
+}
+
+func TestGitlab(t *testing.T) {
```
tamalsaha: Same comments as Github pr

PR #67, opened 2018-02-27
Add test for token auth

nightfury1204:


lib/token_auth_test.go
```diff
@@ -2,20 +2,317 @@ package lib
...
+	"github.com/spf13/afero"
+	auth "k8s.io/api/authentication/v1"
 )
 
-func TestParseCsvFile(t *testing.T) {
-	res, err := LoadTokenFile(runtime.GOPath() + "/src/github.com/appscode/guard/dist/auth/token.csv")
-	if err != nil {
-		t.Error(err)
+func stringArraytoBytes(in []string) []byte {
+	out := ""
```
tamalsaha: []byte(strings.Join(in, "\n")) ?

lib/token_auth_test.go
```diff
@@ -2,20 +2,317 @@ package lib
...
+			"Invalid token",
+			http.StatusUnauthorized,
+			false,
+		},
+	}
+
+	srv := Server{}
+
+	for pos, testData := range dataset {
+		t.Run(fmt.Sprintf("scenario 1 testcase #%v", pos), func(t *testing.T) {
```
tamalsaha: Give it a descriptive name so we know what you are testing here

PR #68, opened 2018-02-28
Add test for azure

nightfury1204:


lib/graph/graph.go
```diff
@@ -203,3 +203,30 @@ func New(clientID, clientSecret, tenantName string) (*UserInfo, error) {
 
 	return u, nil
 }
+
+func NewUserInfo(clientID, clientSecret, tenantName, loginUrl, apiUrl string) (*UserInfo, error) {
```
tamalsaha: If this is only used in test, then this should also be in _test.

lib/azure_test.go
```diff
@@ -1,63 +1,586 @@
...
+	if review.Status.Authenticated {
+		return fmt.Errorf("Expected authenticated false, got true")
+	}
+	if len(review.Status.Error) == 0 {
+		return fmt.Errorf("Expected error message, got empty message")
+	}
+	return nil
+}
+
+func TestCheckAzure(t *testing.T) {
```
tamalsaha: Same comment as github pr

PR #69, opened 2018-03-05
Bug fixes and add CA cert flag for LDAP

nightfury1204:


cmds/installer.go
```diff
@@ -346,7 +378,7 @@ func newService(namespace, addr string) runtime.Object {
 			ClusterIP: host,
 			Ports: []core.ServicePort{
 				{
-					Name:       "api",
+					Name:       "service-api",
```
tamalsaha: Why rename?

cmds/installer.go
```diff
@@ -322,6 +335,25 @@ func newDeployment(opts options) runtime.Object {
...
+			MountPath: "/etc/guard/certs/",
+		}
+		d.Spec.Template.Spec.Containers[0].VolumeMounts = append(d.Spec.Template.Spec.Containers[0].VolumeMounts, volMount)
+
+		vol := core.Volume{
+			Name: "guard-cert",
+			VolumeSource: core.VolumeSource{
+				Secret: &core.SecretVolumeSource{
+					SecretName:  "guard-cert",
+					DefaultMode: types.Int32P(0555),
```
tamalsaha: Why not just `444` ?

cmds/installer.go
```diff
@@ -149,6 +149,19 @@ func NewCmdInstaller() *cobra.Command {
 				buf.WriteString("---\n")
 			}
 
+			if opts.LDAP.CaCertFile != "" {
+				cert, err := ioutil.ReadFile(opts.LDAP.CaCertFile)
+				if err != nil {
+					log.Fatalln(err)
+				}
+				data, err = meta.MarshalToYAML(newSecretForLDAPCert(opts.namespace, map[string][]byte{"ca.crt": cert}), core.SchemeGroupVersion)
```
tamalsaha: Assign the secret to a new variable first. Makes it more readable

lib/server.go
```diff
@@ -101,7 +111,7 @@ func (s Server) ListenAndServe() {
 			tls.TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256,
 			tls.TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256,
 		},
-		ClientAuth: tls.RequireAndVerifyClientCert,
+		ClientAuth: tls.VerifyClientCertIfGiven,
```
tamalsaha: Add a comment that this is needed to pass healthz checks

lib/server.go
```diff
@@ -76,6 +76,16 @@ func (s Server) ListenAndServe() {
...
 	}
 
+	// caCertPool for self signed LDAP sever certificate
+	if s.LDAP.CaCertFile != "" {
+		caCert, err := ioutil.ReadFile(s.LDAP.CaCertFile)
+		if err != nil {
+			log.Fatal(err)
+		}
+		s.LDAP.caCertPool = x509.NewCertPool()
+		s.LDAP.caCertPool.AppendCertsFromPEM(caCert)
```
tamalsaha: Check return value to make sure the cert was found and added

lib/ldap.go
```diff
@@ -109,21 +116,27 @@ func (s Server) checkLDAP(token string) (auth.TokenReview, int) {
...
-		InsecureSkipVerify: s.LDAP.SkipTLSVerification,
-	}
 	var (
 		err  error
 		conn *ldap.Conn
 	)
+
+	tlsConfig := &tls.Config{
+		ServerName:         s.LDAP.ServerAddress,
+		InsecureSkipVerify: s.LDAP.SkipTLSVerification,
```
tamalsaha: Why are we using insecure defaults?

cmds/installer.go
nightfury1204: i renamed it because container port name and service port name is same, but port is different. So, when i testing it in minikube, guard server didn't get any request. In kubernetes logs i found the 'connection refused error'.

tamalsaha: should be fixed now

PR #79, opened 2018-03-11
Ensure max clock skew is no more than 5 sec

tamalsaha:


tamalsaha:
https://github.com/appscode/guard/issues/36

PR #123, opened 2018-04-15
Guard Installation guide for Kubernetes Clusters built with Kubespray

vikas027:
Hello @tamalsaha ,

Please have a look at this doco. I have tried to make it a little detailed so that it is easier to consume even for a newbie.

Cheers,
Vikas


tamalsaha:
Fantastic work @vikas027 ! I have added some minor cleanup.

PR #133, opened 2018-04-30
Add paging to get around directoryObjects.getByIds limit of 1000

amanohar:
When trying to detailed description of AAD membership groups using API getbyids: https://github.com/microsoftgraph/microsoft-graph-docs/blob/master/api-reference/v1.0/api/directoryobject_getbyids.md if AAD member has more than 1000 group memerships

AAD returns:
image

Therefore, guard needs to do page queries to get details of all groups the user belong to. But getbyids does not support query options like $top and $expand and returns error:

    "code": "Request_BadRequest",
    "message": "The following query options are not supported by this request method or cannot be applied to the requested resource: $filter,$expand,$orderby,$count,$inlinecount,$select,$skiptoken,$skip,$top"

So in this case guard cannot rely on AAD to do paging and needs to do paging itself to fetch membership group details from AAD.

Github issue: https://github.com/appscode/guard/issues/132

### DIFF
--- a/auth/providers/azure/graph/graph.go
+++ b/auth/providers/azure/graph/graph.go
@@ -18,9 +18,10 @@ import (
 
 // These are the base URL endpoints for MS graph
 var (
-	baseAPIURL, _ = url.Parse("https://graph.microsoft.com/v1.0")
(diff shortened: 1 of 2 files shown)


amanohar:
@tamalsaha PTAL. UTs pass when run locally.

PR #139, opened 2018-05-11
(testing) Make test synchronous to avoid go routine race condition

nightfury1204:


tamalsaha:
xref: https://github.com/appscode/guard/pull/223

Issue #153, opened 2018-06-19
Azure AAD auth provider is using AAD group's displayName instead of unique objectId for auth


amanohar:
Currently guard updates AAD token of the user with display name of AAD groups the user belongs to. Several AAD groups in an organization can have the same display name. This can cause group claim based auth using Guard webhook to be unreliable.

Code here: https://github.com/appscode/guard/blob/master/auth/providers/azure/graph/graph.go#L212

Fix: Update Guard to use AAD group's unique id/object id `objectId` (which is guaranteed to be unique) instead of display name which might not be unique.

E.g. here: https://msdn.microsoft.com/en-us/library/azure/ad/graph/api/groups-operations

cc: @tamalsaha

tamalsaha:
Fixed in https://github.com/appscode/guard/releases/tag/0.1.4

PR #154, opened 2018-06-20
Allow Azure AAD auth provider to use AAD group ids instead of display name for authn/authz

amanohar:
Allow Azure AAD auth provider to use AAD group ids instead of display name for authentication

This PR will allow Azure AAD auth provider to be set up to use AAD group ids instead of display names for authentication. More details in issue #153


auth/providers/azure/options.go
```diff
@@ -29,6 +30,7 @@ func (o *Options) AddFlags(fs *pflag.FlagSet) {
 	fs.StringVar(&o.ClientID, "azure.client-id", o.ClientID, "MS Graph application client ID to use")
 	fs.StringVar(&o.ClientSecret, "azure.client-secret", o.ClientSecret, "MS Graph application client secret to use")
 	fs.StringVar(&o.TenantID, "azure.tenant-id", o.TenantID, "MS Graph application tenant id to use")
+	fs.BoolVar(&o.UseGroupUID, "azure.use-group-uid", false, "MS Graph application tenant id to use")
```
colemickens: Update the flag usage text, it's copy pasted from the tenant Id.

auth/providers/azure/options.go
```diff
@@ -17,18 +17,21 @@ type Options struct {
 	ClientID     string
 	ClientSecret string
 	TenantID     string
+	UseGroupUID  bool
 }
 
 func NewOptions() Options {
 	return Options{
 		ClientSecret: os.Getenv("AZURE_CLIENT_SECRET"),
+		UseGroupUID: true,
```
tamalsaha: I have changed it to default `true` , since that is the safer option.

amanohar: I left it `false` for backward compatibility reasons so that upgrading existing Guard deployments are not impacted.

Issue #170, opened 2018-07-06
Support B2B auth for Azure provider by supporting both `oid` or `upn` claims in the token


amanohar:
**Scenario:**
User's identity can exist in their organization's AAD tenant (T1) and their AAD application + groups can live in another tenant (T2). 

In this case during device login flow AAD returns a token that's missing UPN: https://docs.microsoft.com/en-us/azure/active-directory/b2b/claims-mapping probably due to: 

Exert from article above:
> For B2B collaboration users, mapping NameID and UPN cross-tenant are prevented for security reasons.

Example of one such token:
```
{
  "aud": "spn:12f61069-cba6-4cc2-dae8-8a7b8ccf41c4",
  "iss": "https://sts.windows.net/<ID_OF_TENANT_T2>/",
  "iat": 1530828265,
  "nbf": 1530828265,
  "exp": 1530832165,
  "acr": "1",
  "aio": "AWQAm/8HAAAASW3Mtsad0xxTH9GT5Cg5Z3OyEoSFjuhrP4OKeZO",
  "amr": [
    "pwd"
  ],
  "appid": "aa2z65ee-9ad4-4645-ab40-95f461599z37",
  "appidacr": "0",
  "email": "<redacted>",
  "family_name": "<redacted>",
  "given_name": "<redacted>",
  "groups": [
    "2a9de113-37eb-4b6c-8de7-7f734146f19b",
    "63ac0747-cd2a-4d31-b38f-9a43bde872b4",
  ],
  "idp": "https://sts.windows.net/<ID_OF_TENANT_T1>/",
  "ipaddr": "31.235.137.123",
  "name": "<redacted>",
  "oid": "<redacted>",
  "scp": "user_impersonation",
  "unique_name": "<redacted>",
  "ver": "1.0"
}
```

**Issue**: 
Currently Guard relies on `upn` to fetch user's group. However, this will only work if the user and the application is in the same AAD tenant.  This is because for B2B collaboration users, mapping NameID and UPN cross-tenant are prevented for security reasons.

**Fix:**
Overall `oid` is a much more reliable way to query user's properties and works cross tenant. Guard can continue using `upn` but provide an option to be setup using `upn` or fallback to using `oid` if `upn` is not present in the token.

tamalsaha:
Fixed by #172

PR #172, opened 2018-07-07
Enable B2B auth for Azure provider by supporting either `oid` or `upn` claim in the token

amanohar:
User's identity can exist in their organization's AAD tenant (T1) and their AAD application + groups can live in another tenant (T2).

This change enables B2B auth scenario where AAD returns a token that's missing `upn` and only contains `oid` so that either `upn` or `oid` claim can be used get user's group memberships.

Issue: #170


tamalsaha:
Thanks @amanohar !

Issue #226, opened 2019-06-27
Guard failed to start because Error: unknown flag: --gitlab.use-group-id


qw1mb0:
After apply YML pod in CrashLoopBackOff state with error:
```
# kubectl -n kube-system  logs guard-bfd4d6559-psx6x
Error: unknown flag: --gitlab.use-group-id
Usage:
  guard run [flags]

```
Guard version:
```
Version = 0.4.0
VersionStrategy = tag
Os = linux
Arch = amd64
CommitHash = 754eb9d99b25bdac7c415c6b987d9290a10afdbe
GitBranch = release-0.4
GitTag = 0.4.0
CommitTimestamp = 2019-02-04T23:24:56
```

nightfury1204:
@qw1mb0 can you check the image version? it shouldn't give this error.

qw1mb0:
`appscode/guard:canary`

nightfury1204:
That's outdated image. Use version taged image: `appscode/guard:0.4.0`

qw1mb0:
So we need to use a specific tag here https://github.com/appscode/guard/blame/master/installer/deployment.go#L47 ?

nightfury1204:
i think it's problem with `guard` binary you have. Try `guard version` to see whether it shows version or not.
You can dowload binary from here https://github.com/appscode/guard/releases.

PR #227, opened 2019-06-27
remove gitlab.use-group-id

qw1mb0:
fixes: #226


nightfury1204:
@qw1mb0 we can not remove this option. see: https://github.com/appscode/guard/issues/181

PR #228, opened 2019-07-02
set installer image to 0.4.0

### LINKED ISSUE #226: Guard failed to start because Error: unknown flag: --gitlab.use-group-id

qw1mb0:
fix #226

### DIFF
--- a/installer/deployment.go
+++ b/installer/deployment.go
@@ -44,7 +44,7 @@ func newDeployment(opts Options) (objects []runtime.Object, err error) {
 					Containers: []core.Container{
 						{
 							Name:  "guard",
-							Image: fmt.Sprintf("%s/guard:%v", opts.PrivateRegistry, stringz.Val(v.Version.Version, "canary")),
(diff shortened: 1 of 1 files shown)


nightfury1204:
I think it's not a problem.  If you see the function `stringz.Val(v.Version.Version, "canary"))`, you will see that it only set `canary` when version is empty. I think main problem is with your guard binary. Check `guard version`.

PR #242, opened 2020-02-03
Add auth provider for firebase.google.com

josefkarasek:
Usage: get your service account from firebase and pass it to guard via
    --firebase.sa-json-file - path to the google service account json file
    Then follow the guide as for other providers.


tamalsaha:
Thanks for the pr, @josefkarasek . Can you please add a doc explaining how to use this auth provider? You can see the existing guides here: https://github.com/appscode/guard/tree/master/docs/guides/authenticator

josefkarasek:
@tamalsaha not sure if getting token from guard applies here, because the origin of the token is Firebase. The client then presents this token to guard and guard checks the token's validity.

PR #245, opened 2020-02-07
Add flag to call Graph api only when overage indicator is present

### LINKED ISSUE #171: Don't query for groups if AAD token already contains groups
**Issue:**
If a user is member of more groups than the overage limit, then Azure AD does not emit groups claim in the token. Instead, it includes an overage claim in the token that indicates to the application to query Graph API to retrieve the user’s group membership. 

However, for users that don't belong to a lot of AAD groups those groups are returned in the token. In that case Guard does not need to make a second call to fetch groups and instead can use groups already returned in the token.

E.g.:
```
{
  "aud": "spn:12f61069-cba6-4cc2-dae8-8a7b8ccf41c4",
  "iss": "https://sts.windows.net/<ID_OF_TENANT_T2>/",
  "iat": 1530828265,
  "nbf": 1530828265,
  "exp": 1530832165,
  "acr": "1",
  "aio": "AWQAm/8HAAAASW3Mtsad0xxTH9GT5Cg5Z3OyEoSFjuhrP4OKeZO",
  "amr": [
    "pwd"
  ],
  "appid": "aa2z65ee-9ad4-4645-ab40-95f461599z37",
  "appidacr": "0",
  "email": "<redacted>",
  "family_name": "<redacted>",
  "given_name": "<redacted>",
  "groups": [
    "2a9de113-37eb-4b6c-8de7-7f734146f19b",
    "63ac0747-cd2a-4d31-b38f-9a43bde872b4",
  ],
  "idp": "https://sts.windows.net/<ID_OF_TENANT_T1>/",
  "ipaddr": "31.235.137.123",
  "name": "<redacted>",
  "oid": "<redacted>",
  "scp": "user_impersonation",
  "unique_name": "<redacted>",
  "ver": "1.0"
}
```

**Fix:** 
Validate to check that groups are present in the token. If they are don't make a second call to Graph for fetching groups.

weinong:
This PR enables an optimization to skip calling to Graph api when any of below conditions satisfies
1. `groups` claim is already present
1. overage claim is not present

It fixes #171

### DIFF
--- a/auth/providers/azure/azure.go
+++ b/auth/providers/azure/azure.go
@@ -160,6 +160,16 @@ func (s Authenticator) Check(token string) (*authv1.UserInfo, error) {
 	if err != nil {
 		return nil, err
 	}
+	if s.Options.ResolveGroupMembershipOnlyOnOverageClaim {
+		groups, skipGraphAPI, err := getGroupsAndCheckOverage(claims)
+		if err != nil {
+			return nil, fmt.Errorf("error in getGroupsAndCheckOverage: %s", err)
+		}
+		if skipGraphAPI {
+			resp.Groups = groups
+			return resp, nil
+		}
+	}
 	if err := s.graphClient.RefreshToken(token); err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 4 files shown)


tamalsaha:
@weinong, can you please request one of your team members to review it ? Please update any doc needed. Same pr is ok.

weinong:
@karataliu PTAL
@tamalsaha doc is updated in this PR

auth/providers/azure/azure.go
```diff
@@ -170,6 +180,85 @@ func (s Authenticator) Check(token string) (*authv1.UserInfo, error) {
...
+		// already maps claim sources to claim names.
+		// however, there is no interface to expose these resolved distributed claims
+		return nil, true, fmt.Errorf("%s is missing in _claim_sources", src)
+	}
+	if ep.Endpoint == "" {
+		// may not be a distributed token
+		return nil, true, nil
+	}
+
+	// return true to proceed to call graph api
```
karataliu: return 'false'?

PR #246, opened 2020-02-07
Add option to verify client ID

### LINKED ISSUE #244: azure auth provider should validate token audience
Current azure auth provider does not validate token audience matching the client ID. [ref](https://github.com/appscode/guard/blob/master/auth/providers/azure/azure.go#L94) and [ref](https://github.com/appscode/guard/blob/d23526bb60dcf4589b2ce45f0420a90342a4ab1d/vendor/github.com/coreos/go-oidc/verify.go#L257). This is not following the [recommended practice](https://docs.microsoft.com/en-us/azure/active-directory/develop/access-tokens#validating-tokens) in validating AAD token:
> To validate access tokens, your app should also validate the issuer, the audience, and the signing tokens.

To prevent from breaking existing guard user, we should add an option to perform client ID validation

weinong:
Add an option in azure auth module to verify client ID which follows the best practice from Azure AD
It fixes #244

### DIFF
--- a/auth/providers/azure/options.go
+++ b/auth/providers/azure/options.go
@@ -45,6 +45,7 @@ type Options struct {
 	AuthMode                                 string
 	AKSTokenURL                              string
 	ResolveGroupMembershipOnlyOnOverageClaim bool
+	VerifyClientID                           bool
 }
 
 func NewOptions() Options {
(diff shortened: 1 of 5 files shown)


weinong:
forced push to update doc. @karataliu PTAL

auth/providers/azure/options.go
```diff
@@ -80,16 +82,16 @@ func (o *Options) Validate() []error {
...
-		}
 	}
 	if o.AuthMode == AKSAuthMode && o.AKSTokenURL == "" {
 		errs = append(errs, errors.New("azure.aks-token-url must be non-empty"))
 	}
 	if o.TenantID == "" {
 		errs = append(errs, errors.New("azure.tenant-id must be non-empty"))
 	}
+	if o.VerifyClientID && o.ClientID == "" {
+		errs = append(errs, errors.New("azure.client-id must be non-empty when azure.verify-client is set"))
```
karataliu: different from those set in options, which is 'azure.verify-clientID' ?

auth/providers/azure/options_test.go
```diff
@@ -63,13 +63,14 @@ var (
...
-			"azure.client-id is empty",
+			"azure.client-id is empty when verify-clientID is true",
 			func(o Options) Options {
 				o.ClientID = empty
+				o.VerifyClientID = true
 				return o
 			},
-			errors.New("azure.client-id must be non-empty"),
-			true,
+			errors.New("azure.client-id must be non-empty when azure.verify-client is set"),
```
karataliu: azure.verify-clientID

weinong:
@tamalsaha PTAL

tamalsaha:
@weinong , do you plan to open any more pr regarding this or should I cut a release?

weinong:
@tamalsaha you can cut the release. The only thing left on my todo is to add metrics. But that will come later. Thank you!

PR #250, opened 2020-05-07
Azure: switch to new graph endpoint for US Government cloud

karataliu:
The graph endpoint in metadata for US Government cloud is in correct, always stick to new endpoint.

Reference:
https://docs.microsoft.com/en-us/graph/deployments#microsoft-graph-and-graph-explorer-service-root-endpoints

### DIFF
--- a/auth/providers/azure/azure.go
+++ b/auth/providers/azure/azure.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/http"
+	"strings"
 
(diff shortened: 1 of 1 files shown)


karataliu:
@tamalsaha This is a bug fix for the specific cloud environment, without this the group claim won't work.
Do you have info when there will be a release including this fix? Thank you.

tamalsaha:
@karataliu , I have released https://github.com/appscode/guard/releases/tag/v0.5.0 . Thanks for your contribution!

Issue #237, opened 2019-12-24
Enhance guard as Authorization Webhook server


krdhruva:
Currently guard is being used in AKS as authentication server for AAD users. In this scenario guard requires to have certain kind of read permission to fetch user information from AAD.
By adding some more read permissions, Guard can be enhanced as Authorization server such that it can verify AAD user access on Kubernetes resources. This will ensure that user doesn't need any additional solution for Authorization.
 
This issue is a feature request to add Authorization webhook server support in Guard for Azure.

tamalsaha:
@krdhruva, are you part of the Azure team? Either way, we would love to have your contribution on this feature.

cc: @amanohar , as contributed previously to Azure feature.

krdhruva:
@tamalsaha, yes I'm part of Azure team.

tamalsaha:
https://github.com/appscode/guard/pull/236 pr coming from your team?

krdhruva:
Yes. @weinong is part of Azure team.

PR #255, opened 2020-05-22
Adding Authz support for Azure

krdhruva:
Adding Authz support for Azure for both AKS and ARC scenarios.
This PR contains cli, code and doc changes. This implements #237


krdhruva:
@weinong , @tamalsaha  can you please take a look?

tamalsaha:
@krdhruva , can you please rerun the go mod command? It seems to fail on CI. We use Go 1.14 currently.

krdhruva:
@tamalsaha we fixed CI code issues. CI is failing in docker login step. It seems it's not able to find docker hub token. Can you please take a look?

krdhruva:
@tamalsaha are we OK to merge?

Makefile
```diff
@@ -325,7 +325,7 @@ lint: $(BUILD_DIRS)
 	    --env HTTP_PROXY=$(HTTP_PROXY)                          \
 	    --env HTTPS_PROXY=$(HTTPS_PROXY)                        \
 	    --env GO111MODULE=on                                    \
-	    --env GOFLAGS="-mod=vendor"                             \
+		--env GOFLAGS="-mod=vendor"                             \
```
weinong, line 328: has additional tab?

authz/types.go
```diff
@@ -0,0 +1,57 @@
...
+)
+
+type orgs []string
+
+var SupportedOrgs orgs
+
+func (o orgs) Has(name string) bool {
+	name = strings.TrimSpace(strings.ToLower(name))
+	for _, org := range o {
+		if org == name {
```
tamalsaha, line 32: Use `strings.EqualFold` instead of `strings.ToLower`

authz/providers/azure/options.go
```diff
@@ -0,0 +1,125 @@
...
+	fs.BoolVar(&o.SkipAuthzForNonAADUsers, "azure.skip-authz-for-non-aad-users", o.SkipAuthzForNonAADUsers, "skip authz for non AAD users")
+}
+
+func (o *Options) Validate(azure azure.Options) []error {
+	var errs []error
+	o.AuthzMode = strings.ToLower(o.AuthzMode)
+	switch o.AuthzMode {
+	case AKSAuthzMode:
+	case ARCAuthzMode:
+	case "":
```
tamalsaha, line 65: So, "" is a valida value? What mode does that default to?

authz/providers/azure/options.go
```diff
@@ -0,0 +1,125 @@
...
+		if azure.ClientSecret == "" {
+			errs = append(errs, errors.New("azure.client-secret must be non-empty"))
+		}
+		if azure.ClientID == "" {
+			errs = append(errs, errors.New("azure.client-id must be non-empty"))
+		}
+	}
+
+	if o.ARMCallLimit > maxPermissibleArmCallLimit {
+		errs = append(errs, errors.New("azure.arm-call-limit must not be more than 4000"))
```
tamalsaha, line 92: fmt.Errorf("azure.arm-call-limit must not be more than %d", maxPermissibleArmCallLimit)

authz/providers/azure/options.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+
+	return errs
+}
+
+func (o Options) Apply(d *apps.Deployment) (extraObjs []runtime.Object, err error) {
+	container := d.Spec.Template.Spec.Containers[0]
```
tamalsaha, line 99: Why not just have the `args` variable? I don't see the container variable is used here.

`args := d.Spec.Template.Spec.Containers[0].Args`

authz/providers/azure/azure.go
```diff
@@ -0,0 +1,126 @@
...
+	c := &Authorizer{}
+
+	authzInfoVal, err := getAuthzInfo(authopts.Environment)
+	if err != nil {
+		return nil, errors.Wrap(err, "Error in getAuthzInfo %s")
+	}
+
+	switch opts.AuthzMode {
+	case ARCAuthzMode:
+		c.rbacClient, err = rbac.New(authopts.ClientID, authopts.ClientSecret, authopts.TenantID, authzInfoVal.AADEndpoint, authzInfoVal.ARMEndPoint, opts.AuthzMode, opts.ResourceId, opts.ARMCallLimit, dataStore, opts.SkipAuthzCheck, opts.AuthzResolveGroupMemberships, opts.SkipAuthzForNonAADUsers)
```
tamalsaha, line 58: Given how many parameters are passed here, I think we should use a `ARCOptions` and `AKSoptions` parameter to pass these values . If you want to go more fancy, functional options can be used. https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis

authz/providers/azure/azure.go
```diff
@@ -0,0 +1,126 @@
...
+	return c, nil
+}
+
+func (s Authorizer) Check(request *authzv1.SubjectAccessReviewSpec) (*authzv1.SubjectAccessReviewStatus, error) {
+	if request == nil {
+		return nil, errors.New("subject access review is nil")
+	}
+
+	// check if user is service account
+	if strings.HasPrefix(strings.ToLower(request.User), "system") {
```
tamalsaha, line 75: We should check against prefix with delimeter `system` -> `system:serviceaccount:`

authz/providers/azure/azure.go
```diff
@@ -0,0 +1,126 @@
...
+		if s.rbacClient.ShouldSkipAuthzCheckForNonAADUsers() {
+			glog.V(3).Infof("Skip RBAC is set for non AAD users. Returning no opinion for user %s. You may observe this for AAD users for 'can-i' requests.", request.User)
+			return &authzv1.SubjectAccessReviewStatus{Allowed: false, Reason: rbac.NoOpinionVerdict}, nil
+		} else {
+			glog.V(3).Infof("Skip RBAC for non AAD user is not set. Returning deny access for non AAD user %s. You may observe this for AAD users for 'can-i' requests.", request.User)
+			return &authzv1.SubjectAccessReviewStatus{Allowed: false, Denied: true, Reason: rbac.NotAllowedForNonAADUsers}, nil
+		}
+	}
+
+	if s.rbacClient.SkipAuthzCheck(request) {
```
tamalsaha, line 90: I think this check should be done first. If user is in this list, does it matter whether this is an AAD user or not?

authz/providers/azure/azure.go
```diff
@@ -0,0 +1,126 @@
...
+		glog.V(3).Infof("user %s is part of skip authz list. returning no op.", request.User)
+		return &authzv1.SubjectAccessReviewStatus{Allowed: false, Reason: rbac.NoOpinionVerdict}, nil
+	}
+
+	exist, result := s.rbacClient.GetResultFromCache(request)
+	if exist {
+		if result {
+			glog.V(3).Infof("cache hit: returning allowed to user")
```
tamalsaha, line 98: `Infof("cache hit: returning allowed to user %s", request.User)` ?

authz/providers/azure/azure.go
```diff
@@ -0,0 +1,126 @@
...
+			return &authzv1.SubjectAccessReviewStatus{Allowed: result, Reason: rbac.AccessAllowedVerdict}, nil
+		} else {
+			glog.V(3).Infof("cache hit: returning denied to user")
```
tamalsaha, line 101: Log user?

authz/providers/azure/azure.go
```diff
@@ -0,0 +1,126 @@
...
+			return &authzv1.SubjectAccessReviewStatus{Allowed: result, Denied: true, Reason: rbac.AccessNotAllowedVerdict}, nil
+		}
+	}
+
+	if s.rbacClient.IsTokenExpired() {
```
tamalsaha, line 106: I think it would be nice, if this could be hiden using something like a `TokenSource` interface found in Go's Oauth2 library. 

- https://github.com/golang/oauth2/blob/master/oauth2.go#L62-L68
- https://github.com/golang/oauth2/blob/master/oauth2.go#L295-L310

This is an idea. Ok, if too complicated to do here.

authz/providers/azure/rbac/rbac.go
```diff
@@ -0,0 +1,259 @@
...
+	"github.com/pkg/errors"
+	authzv1 "k8s.io/api/authorization/v1"
+)
+
+const (
+	managedClusters         = "Microsoft.ContainerService/managedClusters"
+	connectedClusters       = "Microsoft.Kubernetes/connectedClusters"
+	checkAccessPath         = "/providers/Microsoft.Authorization/checkaccess"
+	checkAccessAPIVersion   = "2018-09-01-preview"
+	remaingSubReadARMHeader = "x-ms-ratelimit-remaining-subscription-reads"
```
tamalsaha, line 44: Spelling error `remaing` -> `remaining`

authz/providers/azure/rbac/rbac.go
```diff
@@ -0,0 +1,259 @@
...
+	clusterType              string
+	azureResourceId          string
+	armCallLimit             int
+	dataStore                authz.Store
+	skipCheck                map[string]void
+	retrieveGroupMemberships bool
+	skipAuthzForNonAADUsers  bool
+}
+
+func newAccessInfo(tokenProvider graph.TokenProvider, rbacURL *url.URL, clsuterType, resourceId string, armCallLimit int, dataStore authz.Store, skipList []string, retrieveGroupMemberships, skipAuthzForNonAADUsers bool) (*AccessInfo, error) {
```
tamalsaha, line 68: `clsuterType` -> `clusterType` ?

authz/providers/azure/rbac/rbac.go
```diff
@@ -0,0 +1,259 @@
...
+	u := &AccessInfo{
```
tamalsaha, line 69: I think functional options will be a good idea here. https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis

authz/providers/azure/rbac/rbac.go
```diff
@@ -0,0 +1,259 @@
...
+		headers: http.Header{
+			"Content-Type": []string{"application/json"},
+		},
+		apiURL:                   rbacURL,
+		tokenProvider:            tokenProvider,
+		azureResourceId:          resourceId,
+		armCallLimit:             armCallLimit,
+		dataStore:                dataStore,
+		retrieveGroupMemberships: retrieveGroupMemberships,
+		skipAuthzForNonAADUsers:  skipAuthzForNonAADUsers}
```
tamalsaha, line 80: Nit: Move the closing curly brace to newline.

authz/providers/azure/rbac/rbac.go
```diff
@@ -0,0 +1,259 @@
...
+	expiresAt time.Time
+	// These allow us to mock out the URL for testing
+	apiURL *url.URL
+
+	tokenProvider            graph.TokenProvider
+	clusterType              string
+	azureResourceId          string
+	armCallLimit             int
+	dataStore                authz.Store
+	skipCheck                map[string]void
```
tamalsaha, line 63: We could also use `sets.String` https://github.com/kubernetes/apimachinery/blob/v0.18.3/pkg/util/sets/string.go#L26

authz/providers/azure/rbac/rbac.go
```diff
@@ -0,0 +1,259 @@
...
+// AccessInfo allows you to check user access from MS RBAC
+type AccessInfo struct {
+	headers   http.Header
+	client    *http.Client
+	expiresAt time.Time
+	// These allow us to mock out the URL for testing
+	apiURL *url.URL
+
+	tokenProvider            graph.TokenProvider
+	clusterType              string
```
tamalsaha, line 59: I think we could define a string type

```
type ClusterType string

const (
    ARCCluster ClusterType = "arc"
    AKSCluster ClusterType = "aks"
)

```

authz/providers/azure/rbac/checkaccessreqhelper.go
```diff
@@ -0,0 +1,338 @@
...
+	}
+	return finalGroups
+}
+
+func getActionName(verb string) string {
+	/* special verbs
+	use verb on podsecuritypolicies resources in the policy API group
+	bind and escalate verbs on roles and clusterroles resources in the rbac.authorization.k8s.io API group
+	impersonate verb on users, groups, and serviceaccounts in the core API group
+	userextras in the authentication.k8s.io API group
```
tamalsaha, line 153: How is `userextras` relevant here?

authz/providers/azure/rbac/rbac.go
```diff
@@ -0,0 +1,259 @@
...
+func (a *AccessInfo) GetResultFromCache(request *authzv1.SubjectAccessReviewSpec) (bool, bool) {
+	var result bool
+	key := getResultCacheKey(request)
+	glog.V(10).Infof("Cache search for key: %s", key)
+	found, _ := a.dataStore.Get(key, &result)
+	return found, result
+}
+
+func (a *AccessInfo) SkipAuthzCheck(request *authzv1.SubjectAccessReviewSpec) bool {
+	if a.clusterType == connectedClusters {
```
tamalsaha, line 160: I am confused by this check. To me it seems that a.clusterType can be either "arc" or "aks"

authz/providers/azure/rbac/checkaccessreqhelper.go
```diff
@@ -0,0 +1,338 @@
...
+		authInfo.AuthorizationEntity.Id = path.Join(authInfo.AuthorizationEntity.Id, subRevReq.NonResourceAttributes.Path, getActionName(subRevReq.NonResourceAttributes.Verb))
+	}
+	return authInfo
+}
+
+func getResultCacheKey(subRevReq *authzv1.SubjectAccessReviewSpec) string {
+	cacheKey := subRevReq.User
+
+	if subRevReq.ResourceAttributes != nil {
+		if subRevReq.ResourceAttributes.Namespace != "" {
```
tamalsaha, line 210: I think this can be tricked by creating a namespace that matches a group. We should always set a path component even for empty ns or group.

```
	cacheKey = path.Join(cacheKey, defaultDir(subRevReq.ResourceAttributes.Namespace))
	cacheKey = path.Join(cacheKey, defaultDir(subRevReq.ResourceAttributes.Group))


func defaultDir(s string) string {
	if s != "" {
		return s
	}
	return "-" // invalid for a namespace
}
```

authz/providers/azure/rbac/checkaccessreqhelper.go
```diff
@@ -0,0 +1,338 @@
...
+	)
+	err := json.Unmarshal(body, &response)
+	if err != nil {
+		glog.V(10).Infof("Failed to parse checkacccess response. Error:%s", err.Error())
+		return nil, errors.Wrap(err, "Error in unmarshalling check access response.")
+	}
+
+	if glog.V(10) {
+		binaryData, _ := json.MarshalIndent(response, "", "    ")
+		glog.Infof("check access response:%s", binaryData)
```
tamalsaha, line 325: `glog.Infof` -> `glog.V(10).Infof`

authz/providers/azure/rbac/checkaccessreqhelper.go
```diff
@@ -0,0 +1,338 @@
...
+		denied   bool
+		verdict  string
+	)
+	err := json.Unmarshal(body, &response)
+	if err != nil {
+		glog.V(10).Infof("Failed to parse checkacccess response. Error:%s", err.Error())
+		return nil, errors.Wrap(err, "Error in unmarshalling check access response.")
+	}
+
+	if glog.V(10) {
```
tamalsaha, line 323: Should this be done before `json.Unmarshal` ?

authz/providers/azure/rbac/rbac.go
```diff
@@ -0,0 +1,259 @@
...
+
+func (a *AccessInfo) RefreshToken() error {
+	resp, err := a.tokenProvider.Acquire("")
+	if err != nil {
+		glog.Errorf("%s failed to refresh token : %s", a.tokenProvider.Name(), err.Error())
+		return errors.Wrap(err, "failed to refresh rbac token")
+	}
+
+	// Set the authorization headers for future requests
+	a.headers.Set("Authorization", fmt.Sprintf("Bearer %s", resp.Token))
```
tamalsaha, line 132: This call is not thread-safe as it just uses maps underneath. We need a mutex to avoid getting concurrent access panics.

server/authzhandler.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+
+	resp, err := client.Check(&data.Spec)
+	writeAuthzResponse(w, &data.Spec, resp, err)
+}
+
+func (s *Authzhandler) getAuthzProviderClient(org string) (authz.Interface, error) {
+	switch strings.ToLower(org) {
+	case azure.OrgType:
+		return azure.New(s.AuthzRecommendedOptions.Azure, s.AuthRecommendedOptions.Azure, s.Store)
```
tamalsaha, line 77: I think we can do better by reusing the rbac client across requests. This will allow re-using the token . What do you think?

authz/providers/azure/rbac/checkaccessreqhelper.go
krdhruva: As we have taken dependency on Extra filed from Authentication stack which is mandatory for validating user access on Azure, user is required to pass userextras in kubeconfig if he wants to use impersonate.

authz/providers/azure/rbac/rbac.go
```diff
@@ -0,0 +1,259 @@
...
+		return nil, errors.Errorf("request %s failed with status code: %d and response: %s", req.URL.Path, resp.StatusCode, string(data))
+	} else {
+		remaining := resp.Header.Get(remaingSubReadARMHeader)
+		glog.Infof("Remaining request count in ARM instance:%s", remaining)
+		count, _ := strconv.Atoi(remaining)
+		if count < a.armCallLimit {
+			if glog.V(10) {
+				glog.V(10).Infoln("Closing idle TCP connections.")
+			}
+			// Usually ARM connections are cached by destinatio ip and port
```
tamalsaha, line 244: `destinatio` -> `destination`

authz/providers/azure/rbac/checkaccessreqhelper.go
tamalsaha: Ok. `userextras` is not a verb just the extra info `oid`. I was confused because it was listed under the `special verbs` headline. May be move it to the same line as `impersonate` verb?

tamalsaha:
@krdhruva , I am done with the review. Please let me know what you think.

authz/providers/azure/rbac/checkaccessreqhelper.go
krdhruva: Yes, userextras would be resource and will be handled by impersonate. It is not much clear from the comment. Will update the comment.

authz/providers/azure/azure.go
krdhruva: Although comment mentions service account, idea here is to return 'no opinion' for all system accounts like system:node, system:kube-scheduler etc along with service accounts. I will update the comment and log statement.

authz/providers/azure/azure.go
krdhruva: During testing I have observed that no. of request containing service accounts and system accounts are higher than AAD users. Hence its ideal to check for system accounts before checking in the list. I can add this before cert users though.

authz/providers/azure/azure.go
krdhruva: Fixed by passing auth and authz opts.

authz/providers/azure/azure.go
krdhruva: I have skipped this.

authz/providers/azure/rbac/rbac.go
krdhruva: I think there won't be any perf improvement with this. Hence skipping this.

authz/providers/azure/rbac/rbac.go
krdhruva: Due to other comment fix, this is not required now.

authz/providers/azure/rbac/checkaccessreqhelper.go
krdhruva: This was redundant hence removed this.

server/authzhandler.go
krdhruva: Agreed. Made rbac client singleton.

krdhruva:
@tamalsaha I have incorporated all review comments and updated PR.

krdhruva:
@tamalsaha can you please take a look?

authz/providers/azure/azure.go
tamalsaha: Sounds good. Please add the delimiter `:` to make sure it is checked correctly.

`strings.HasPrefix(strings.ToLower(request.User), "system:")`

authz/providers/azure/rbac/rbac.go
```diff
@@ -0,0 +1,280 @@
...
+func (a *AccessInfo) SetResultInCache(request *authzv1.SubjectAccessReviewSpec, result bool, store authz.Store) error {
+	key := getResultCacheKey(request)
+	glog.V(10).Infof("Cache set for key: %s, value: %t", key, result)
+	return store.Set(key, result)
+}
+
+func (a *AccessInfo) AllowNonResPathDiscoveryAccess(request *authzv1.SubjectAccessReviewSpec) bool {
+	if request.NonResourceAttributes != nil && a.allowNonResDiscoveryPathAccess && strings.EqualFold(request.NonResourceAttributes.Verb, "get") {
+		path := strings.ToLower(request.NonResourceAttributes.Path)
+		if strings.HasPrefix(path, "/api") || strings.HasPrefix(path, "/openapi") || strings.HasPrefix(path, "/version") || strings.HasPrefix(path, "/healthz") {
```
tamalsaha, line 193: - What about `/apis` ?

```
 kubectl get clusterrole system:discovery -o yaml
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  annotations:
    rbac.authorization.kubernetes.io/autoupdate: "true"
  creationTimestamp: "2020-05-31T20:46:20Z"
  labels:
    kubernetes.io/bootstrapping: rbac-defaults
  name: system:discovery
  resourceVersion: "43"
  selfLink: /apis/rbac.authorization.k8s.io/v1/clusterroles/system%3Adiscovery
  uid: c8118371-a37f-11ea-85c5-0242ac120002
rules:
- nonResourceURLs:
  - /api
  - /api/*
  - /apis
  - /apis/*
  - /healthz
  - /openapi
  - /openapi/*
  - /version
  - /version/
  verbs:
  - get
```

- Also, prefix checks should use a delimiter to make sure we are actually matching against a path component. So, these checks should match prefix against `/api/`, `/apis/` to accidentally matching against `/api123`, etc.

authz/providers/azure/azure.go
```diff
@@ -0,0 +1,150 @@
...
+			glog.V(3).Infof("cache hit: returning allowed to user %s", request.User)
+			return &authzv1.SubjectAccessReviewStatus{Allowed: result, Reason: rbac.AccessAllowedVerdict}, nil
+		} else {
+			glog.V(3).Infof("cache hit: returning denied to user %s", request.User)
+			return &authzv1.SubjectAccessReviewStatus{Allowed: result, Denied: true, Reason: rbac.AccessNotAllowedVerdict}, nil
+		}
+	}
+
+	// if set true, webhook will allow access to discovery APIs for authenticated users. If false, access check will be performed on Azure.
+	if s.rbacClient.AllowNonResPathDiscoveryAccess(request) {
```
tamalsaha, line 116: Is this necessary because Kubernetes automatically creates `system:discovery` cluster role and cluster role binding to give this permission to any authenticated users.

authz/providers/azure/rbac/rbac.go
```diff
@@ -0,0 +1,280 @@
...
+		expIn := time.Duration(resp.Expires) * time.Second
+		a.expiresAt = time.Now().Add(expIn - expiryDelta)
+		glog.Infof("Token refreshed successfully on %s. Expire at:%s", time.Now(), a.expiresAt)
+	}
+
+	return nil
+}
+
+func (a *AccessInfo) IsTokenExpired() bool {
+	if a.expiresAt.Before(time.Now()) {
```
tamalsaha, line 157: Nit: Could be simplified to 

` return a.expiresAt.Before(time.Now())`

authz/providers/azure/rbac/rbac.go
```diff
@@ -0,0 +1,280 @@
...
+		glog.V(10).Infof("checkAccessURI:%s", checkAccessURL.String())
+		glog.V(10).Infof("binary data:%s", binaryData)
+	}
+
+	req, err := http.NewRequest(http.MethodPost, checkAccessURL.String(), buf)
+	if err != nil {
+		return nil, errors.Wrap(err, "error creating check access request")
+	}
+	// Set the auth headers for the request
+	req.Header = a.headers
```
tamalsaha, line 236: I think you need mutex.Lock() to read this map to avoid concurrent access to this map.

authz/providers/azure/azure.go
krdhruva: webhook will be enabled before RBAC and we need to handle this. As Kubernetes automatically adds all authenticated users to system:discovery, we are good to allow access on these paths. Hence I have added this check which would be true by default.

authz/providers/azure/rbac/rbac.go
krdhruva: Prefix /api covers all 4 cases. What I have observed is if user creates request on nonexistent non-res path /api123, request returns the right set of non-res paths as below:

http://127.0.0.1:8003/api123
{
  "paths": [
    "/apis",
    "/apis/",
    "/apis/apiextensions.k8s.io",
    "/apis/apiextensions.k8s.io/v1beta1",
    "/healthz",
    "/healthz/etcd",
    "/healthz/log",
    "/healthz/ping",
    "/healthz/poststarthook/crd-informer-synced",
    "/healthz/poststarthook/generic-apiserver-start-informers",
    "/healthz/poststarthook/start-apiextensions-controllers",
    "/healthz/poststarthook/start-apiextensions-informers",
    "/metrics",
    "/openapi/v2",
    "/version"
  ]

This behavior is consistent with K8s RBAC. Hence IMHO there should not be any issue in accidently matching against nonexistent non-res path. Do you see any escalation of privilege here?

tamalsaha: Using a delimiter for prefix checks just safer. But I think it is ok for current paths.

krdhruva:
@tamalsaha I have resolved all the comments and updated PR. Kindly have a look.

authz/providers/azure/rbac/rbac.go
```diff
@@ -0,0 +1,283 @@
...
+
+	tokenProvider                  graph.TokenProvider
+	clusterType                    string
+	azureResourceId                string
+	armCallLimit                   int
+	skipCheck                      map[string]void
+	retrieveGroupMemberships       bool
+	skipAuthzForNonAADUsers        bool
+	allowNonResDiscoveryPathAccess bool
+	lock                           *sync.RWMutex
```
tamalsaha, line 75: This can be just `lock sync.RWMutex`. The only requirement is that `AccessInfo` is always passed as pointer which is what you have done.

authz/providers/azure/rbac/rbac.go
```diff
@@ -0,0 +1,283 @@
...
+		}
+	}
+	return false
+}
+
+func (a *AccessInfo) setReqHeaders(req *http.Request) {
+	a.lock.RLock()
+	defer a.lock.RUnlock()
+	// Set the auth headers for the request
+	req.Header = a.headers
```
tamalsaha, line 200: I am not sure this is enough. In Go, maps are pointer types. So, req.Header is still pointing to the same map. I think you need to make a copy.

```
if req.Header == nil {
	req.Header = make(http.Header)
}
for k, values := range a.headers {
  ...
}

```

tamalsaha:
Thanks a lot, @krdhruva for addressing my comments. Do you want me to cut a release?

krdhruva:
Thanks a lot @tamalsaha for reviewing and merging my PR. Yes, please release a new version with the changes.

tamalsaha:
Done ! https://github.com/appscode/guard/releases/tag/v0.6.0-rc.0

PR #259, opened 2020-06-05
Changed azure authz support to use v1beta1 version instead of v1 for SubjectAccessReview

Anumita:
Making this change as even though we are expecting v1 version of SAR, we are still getting v1beta1 version. Also, in v1beta1 version there is k8s bug where instead of "groups" SAR decodes it as "group". Even though this is fixed in v1, we are not being able to use v1 version. Will be raising a bug on k8s on this regarding how we can use v1 version. Till then as we don't want to explicity call graph every time to get the groups, we are changing to use v1beta1 version so that we can get the groups the user is part of

Signed-off-by: Anumita Shenoy <ansheno@microsoft.com>

### DIFF
--- a/authz/providers/azure/azure.go
+++ b/authz/providers/azure/azure.go
@@ -27,7 +27,7 @@ import (
 	"github.com/Azure/go-autorest/autorest/azure"
 	"github.com/golang/glog"
 	"github.com/pkg/errors"
-	authzv1 "k8s.io/api/authorization/v1"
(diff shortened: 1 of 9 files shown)


Anumita:
K8s bug: https://github.com/kubernetes/kubernetes/issues/91698

Anumita:
@weinong , @tamalsaha , could you review this PR?

weinong:
relate to https://github.com/kubernetes/kubernetes/pull/85441

PR #288, opened 2020-10-21
Add proxy support in Guard (#20)

Anumita:
Added proxy support for guard and additional metrics for authz

Proxy Support
Added 4 parameters for proxy in the installer command
1. 3 are for the common proxy environment variables (HTTP_PROXY, HTTPS_PROXY, NO_PROXY) -  "guard-proxy" secret (installer/secrets.go) is created which is loaded as environment variable in the deployment.yaml. 
     --proxy-http, --proxy-https, --proxy-skip-range
2. --proxy-cert is for proxy with cert authentication. For this, user provides the path to their proxy cert to the --proxy-cert args. Internally, we read the data from the file and create another secret ("guard-proxy-cert"). For proxy with cert, the cert should be part of "/etc/ssl/certs/ca-certificates.pem". This is achieved by running update-ca-certificates command. The update-ca-certificates command takes the certs present in /usr/local/share/ca-certificates and adds them to  /etc/ssl/certs/ca-certificates.pem. To achieve this, we are doing the following:
    -  create two volumes , one is empty directory volume called "ssl-certs" and the other is a volume created from the "guard-proxy-cert" secret called "proxy-certstore". 
    -  An init container (image: nginx:stable-alpine (9 mb)) which does the job of running the update-ca-certificates command. To this container we mount the "ssl-certs" volume at path "/etc/ssl/certs" and the "proxy-certstore" volume at "/usr/local/share/ca-certificates" path. The container will run update-ca-certificates and create the ca-certificates.pem file in the /etc/ssl/certs directory which is from the volume.
   - The main "guard" container has the "ssl-certs" volume mounted at /etc/ssl/certs path. As the volume is shared between the init container and guard container, when the init container updates the /etc/ssl/certs, the guard container will also have the updated ca-certificates.pem file which contains the proxy cert. This way the http client used to make the requests will be able to do successfully through proxy.

The proxy support is for azure-arc scenario where users will be setting up guard by themselves and also in future when arc switches to automating the guard setup for customers

### DIFF
--- a/auth/providers/azure/graph/aks_tokenprovider.go
+++ b/auth/providers/azure/graph/aks_tokenprovider.go
@@ -33,9 +33,12 @@ type aksTokenProvider struct {
 
 // NewAKSTokenProvider returns a TokenProvider that implements On-Behalf-Of flow using AKS first party service
 func NewAKSTokenProvider(tokenURL, tenantID string) TokenProvider {
+	tr := &http.Transport{
+		Proxy: http.ProxyFromEnvironment,
+	}
 	return &aksTokenProvider{
 		name:     "AKSTokenProvider",
-		client:   &http.Client{},
+		client:   &http.Client{Transport: tr},
 		tokenURL: tokenURL,
 		tenantID: tenantID,
 	}
--- a/auth/providers/azure/graph/clientcredential_tokenprovider.go
+++ b/auth/providers/azure/graph/clientcredential_tokenprovider.go
@@ -39,9 +39,12 @@ type clientCredentialTokenProvider struct {
 // NewClientCredentialTokenProvider returns a TokenProvider that implements OAuth client credential flow on Azure Active Directory
 // https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-oauth2-client-creds-grant-flow#get-a-token
 func NewClientCredentialTokenProvider(clientID, clientSecret, loginURL, scope string) TokenProvider {
+	tr := &http.Transport{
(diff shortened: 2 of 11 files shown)


Anumita:
@tamalsaha , the image i have chosen nginx:stable-alpine [()](https://hub.docker.com/_/nginx?tab=tags&page=1&name=stable-alpine) has support for all the archs that guard supports. 

Just had one thing to clarify here , should we add a validation for the args that i have added , that they should be used only when the provider is azure, as i not sure whether this will work for other providers?

@krdhruva has already reviewed and approved this pr in my forked repo.

Anumita:
hey @tamalsaha , gentle reminder

tamalsaha:
Sorry for the delay. I am going to take a look by soon.

Anumita:
hey, @tamalsaha , gentle reminder :). Could you take a look at https://github.com/appscode/guard/pull/293 as well?

Anumita:
hey @tamalsaha , could you please review as this is a blocker for us for release?

tamalsaha (approved):

tamalsaha:
> Just had one thing to clarify here , should we add a validation for the args that i have added , that they should be used only when the provider is azure, as i not sure whether this will work for other providers?

Let's keep it for everyone. The PROXY env vars are pretty generic.

tamalsaha:
Sorry for the delay. It has been one hell of week.

Anumita:
np, thanks! 👍

Issue #320, opened 2021-04-15
Non interactive logins - Azure provider 


dany74q:
Hey hey !

In the past few days, I've had a chance to look over the AKS-AAD integrations (which utilize guard, under the hood);
That was during some research I'm doing for implementing non interactive logins for AAD enabled clusters (both legacy & managed). 

Specifically, I'm looking at auth via service principals, passing a client_credentials-flow token, which holds a service principals object id claim and no UPN claim. 

Managed AAD clusters have it mostly solved (a-la kubelogin) - one only needs to issue a token for the multi tenant AKS server app, and any directory entity could be used in this flow - the groups JWT claim is considered along with any overage data. 

However, in case of overage - ms graph is consulted, and the request fails for SPNs, as msgraph 404-s fetching group memberships for service principals;
This is because the graph API currently used only supports retrieving memberships for users (and not service principals).  

As for legacy clusters, the groups JWT claim isn't considered at all -  
ms graph is always consulted in fetching group memberships for the given object id;
Specifically, the "/users/id/getMemberGroups" endpoint is used. 

When an spn oid is passed - the API above 404-s in the same manner and it fails the request altogether. 

I believe that ms graph had no API for retrieving groups for any given entity back in the day, but now one does exist -"/directoryObjects/id/getMemberGroups". 

I was wondering if it would be a good idea to migrate to the new endpoint - this would enable non interactive login flows to legacy clusters and fix the flow for managed integrations for SPNs assigned to many groups. 

Otherwise, it might make sense to have the ms graph call be best effort - returning blank groups in cases of error;
It's a bit unfortunate that not being able to retrieve groups, fails the auth attempt altogether - 
when reaching that code path, we have a verified JWT at hand with some object id,
it might have made sense to pass it onward and check for any direct k8s role mapping. 

Another suggestion might be to flip the flag which considers the given groups claim (on AKS side) for legacy clusters - closing the disparity between the two integrations. 
 

Would love to hear your two cents on this @weinong

Thanks !

dany74q:
Hey @weinong !

As discussed in our short email exchange - this also happens in the managed AAD integration for SPN-s w/ group overage;
In the flow, the SPN oid is still passed to the graph API - it 404s (as it's not a user-entity) and that fails the request.

I think it would be great to migrate to the new (now GA) API that can list group memberships for both SPNs and users:
https://docs.microsoft.com/en-us/graph/api/directoryobject-getmembergroups?view=graph-rest-1.0&tabs=http

[...]

PR #321, opened 2021-04-21
Migrated to new directoryObjects getGroupsById graph API

dany74q:
Fixes: https://github.com/kubeguard/guard/issues/320

azure.go
- A userInfo wrapper that contains the oid and upn claims was introduced

azure.md
- Updated docs to include the new Application.Read.All permission

azure_test.go
- Migrated oid tests to new endpoint

graph.go
- Extracted out getMemberGroupsGraphURL
- Will now use the new directory objects endpoint for object IDs
- Will fallback to the previous endpoint for UPNs (the new one does not support UPNs)
- Will now use Url.Parse to retrieve a relative URL - it handles escaping (the UPN may contain '#' which need to be url-escaped)
- Fixed warning

graph_test.go
- Added TestGetMemberGroupsGraphURL

obo-server-app.png
- Updated with new Application.Read.All permission


dany74q:
Tests pass - there's something off w/ the docker login step right after (didn't touch anything in that region though)

dany74q:
Closing per our discussion in https://github.com/kubeguard/guard/issues/320

PR #325, opened 2021-07-22
Add support for pod/exec rbac

Anumita:
For pod/exec action, kube-apiserver sends two sar requests, one for pods/read and then for pods/exec/create. With this change, user will be able to only give/block exec into a pod.

Also added log line to make sure we know when a SAR request comes to guard. We are logging the request body here


server/authzhandler.go
```diff
@@ -22,6 +22,7 @@ import (
 	"go.kubeguard.dev/guard/authz"
 	"go.kubeguard.dev/guard/authz/providers/azure"
 
+	"github.com/golang/glog"
```
tamalsaha, line 25: We are only using `k8s.io/klog/v2` and avoid `glog`. So, please use that.

Anumita: removed glog

authz/providers/azure/rbac/checkaccessreqhelper.go
```diff
@@ -195,6 +195,17 @@ func getActionName(verb string) string {
 	}
 }
 
+func getResourceAndAction(subRevReq *authzv1.SubjectAccessReviewSpec) string {
```
krdhruva, line 198: should we add/update existing unit tests around this area?

server/authzhandler.go
```diff
@@ -33,6 +34,7 @@ type Authzhandler struct {
 }
 
 func (s *Authzhandler) ServeHTTP(w http.ResponseWriter, req *http.Request) {
+	klog.Infof("Recieved subject access review request")
```
krdhruva, line 37: let's add req.body in log statement to identify individual request.

krdhruva (changes requested):
⏲️

krdhruva (dismissed):

server/authzhandler.go
```diff
@@ -33,6 +35,13 @@ type Authzhandler struct {
 }
 
 func (s *Authzhandler) ServeHTTP(w http.ResponseWriter, req *http.Request) {
+	body, err := ioutil.ReadAll(req.Body)
```
weinong, line 38: is req.Body being closed?

https://pkg.go.dev/net/http#pkg-overview

The client must close the response body when finished with it:
```go
resp, err := http.Get("http://example.com/")
if err != nil {
	// handle error
}
defer resp.Body.Close()
body, err := io.ReadAll(resp.Body)
// ...
```

weinong (approved):
LGTM. Thanks!

Anumita:
@tamalsaha could you please release the image containing this commit? We can release the image on our side only if it has the same tag as the image released here😅

tamalsaha:
Done! https://github.com/kubeguard/guard/releases/tag/v0.8.0

Anumita:
> Done! https://github.com/kubeguard/guard/releases/tag/v0.8.0

Thanks a lot!

PR #326, opened 2021-08-25
Fixed exec data action cache issue

Anumita:
Wrond action was getting saved in the cache

Signed-off-by: Anumita <ansheno@microsoft.com>


krdhruva (approved):

Anumita:
@tamalsaha , could you please release an image containing this commit today if possible? It's a fix for a bug hence the urgency :). Maybe we can release it as a patch version (0.8.1)?

PR #339, opened 2022-03-21
Added support for new Namespace scope format

Anumita:
This is for adding support the new namespace scope for authorization in guard.
Current namespace scope: namespaces/kube-system
Added useNamespaceResourceScopeFormat which is false by default
On true, scope used will be providers/Microsoft.KubernetesConfiguration/namespaces


authz/providers/azure/rbac/checkaccessreqhelper.go
```diff
@@ -245,7 +250,7 @@ func getResultCacheKey(subRevReq *authzv1.SubjectAccessReviewSpec) string {
 	return cacheKey
 }
 
-func prepareCheckAccessRequestBody(req *authzv1.SubjectAccessReviewSpec, clusterType, resourceId string) (*CheckAccessRequest, error) {
+func prepareCheckAccessRequestBody(req *authzv1.SubjectAccessReviewSpec, clusterType, resourceId string, useNamespaceResourceScopeFormat bool) (*CheckAccessRequest, error) {
```
krdhruva, line 253: Pls add unit test to validate resource Id string by passing ns attribute for both ns format.

krdhruva (approved):

weinong (approved):
LGTM

PR #343, opened 2022-09-16
Increase read time out for Guard

krdhruva:
We have been observing various client connection errors due to default values in http client & server.

stream error: stream ID 1; INTERNAL_ERROR; received from peer -> due to aggressive read time out
connect: cannot assign requested address -> due to default MaxIdleConnsPerHost.

Raising PR to fix these by making read time out value configurable.


weinong (approved):
LGTM

tamalsaha:
Released https://github.com/kubeguard/guard/releases/tag/v0.11.1

PR #350, opened 2023-01-15
Add context timeout for checkaccess requests and fix metrics

Anumita:
Context timeout - 
Change to using https://pkg.go.dev/golang.org/x/sync/errgroup which can propagate error , it is a wrapper on waitgroups. Added a context timeout of 23 seconds.

Added contexttimeout metric as well

Discover Resources metrics:
Added metrics for  - 
1. Total duration for discovering resources which includes both apiserver and get operations call
2. Duration for apiserver call
3. Duration for get operations call

Fix metrics:
1. SAR status was returning 200 code regardless of whether there were any errors or not. Utilized existing withCode struct to make sure we send an appropriate errorcode. The divisions of errorcode are:
a. if checkaccess fails , we will send back errorcode which will be the response status code.
b. if it's any other error it will either be statusbadrequest if client related or statusInternalServerError otherwise

7. Fixed checkaccess requests total and failed metrics to included statuscode as a dimension. This will help us get the success rate
8. Added metrics for checkaccess latency as well

### DIFF
--- a/authz/providers/azure/azure_test.go
+++ b/authz/providers/azure/azure_test.go
@@ -29,6 +29,7 @@ import (
 	authzOpts "go.kubeguard.dev/guard/authz/providers/azure/options"
 	"go.kubeguard.dev/guard/authz/providers/azure/rbac"
 	azureutils "go.kubeguard.dev/guard/util/azure"
+	errutils "go.kubeguard.dev/guard/util/error"
 
 	"github.com/appscode/pat"
 	"github.com/stretchr/testify/assert"
@@ -147,6 +148,9 @@ func TestCheck(t *testing.T) {
 		assert.NotNil(t, resp)
 		assert.Equal(t, resp.Allowed, true)
 		assert.Equal(t, resp.Denied, false)
+		if v, ok := err.(errutils.HttpStatusCode); ok {
+			assert.Equal(t, v.Code(), http.StatusOK)
+		}
 	})
 
 	t.Run("unsuccessful request", func(t *testing.T) {
@@ -167,6 +171,9 @@ func TestCheck(t *testing.T) {
 		assert.Nilf(t, resp, "response should be nil")
 		assert.NotNilf(t, err, "should get error")
 		assert.Contains(t, err.Error(), "Error occured during authorization check")
+		if v, ok := err.(errutils.HttpStatusCode); ok {
+			assert.Equal(t, v.Code(), http.StatusInternalServerError)
+		}
 	})
 
 	t.Run("context timeout request", func(t *testing.T) {
@@ -187,5 +194,8 @@ func TestCheck(t *testing.T) {
 		assert.Nilf(t, resp, "response should be nil")
 		assert.NotNilf(t, err, "should get error")
 		assert.Contains(t, err.Error(), "Checkaccess requests have timed out")
+		if v, ok := err.(errutils.HttpStatusCode); ok {
+			assert.Equal(t, v.Code(), http.StatusInternalServerError)
+		}
 	})
 }
--- a/authz/providers/azure/azure_test.go
+++ b/authz/providers/azure/azure_test.go
@@ -71,7 +71,7 @@ func clientSetup(serverUrl, mode string) (*Authorizer, error) {
 	return c, nil
 }
 
-func serverSetup(loginResp, checkaccessResp string, loginStatus, checkaccessStatus int) (*httptest.Server, error) {
+func serverSetup(loginResp, checkaccessResp string, loginStatus, checkaccessStatus int, sleepFor time.Duration) (*httptest.Server, error) {
 	listener, err := net.Listen("tcp", "127.0.0.1:")
 	if err != nil {
 		return nil, err
@@ -85,6 +85,7 @@ func serverSetup(loginResp, checkaccessResp string, loginStatus, checkaccessStat
 	}))
 
 	m.Post("/arm/", http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		time.Sleep(sleepFor)
 		w.WriteHeader(checkaccessStatus)
 		_, _ = w.Write([]byte(checkaccessResp))
 	}))
@@ -98,8 +99,8 @@ func serverSetup(loginResp, checkaccessResp string, loginStatus, checkaccessStat
 	return srv, nil
 }
 
-func getServerAndClient(t *testing.T, loginResp, checkaccessResp string, checkaccessStatus int) (*httptest.Server, *Authorizer, authz.Store) {
-	srv, err := serverSetup(loginResp, checkaccessResp, http.StatusOK, checkaccessStatus)
+func getServerAndClient(t *testing.T, loginResp, checkaccessResp string, checkaccessStatus int, sleepFor time.Duration) (*httptest.Server, *Authorizer, authz.Store) {
+	srv, err := serverSetup(loginResp, checkaccessResp, http.StatusOK, checkaccessStatus, sleepFor)
 	if err != nil {
 		t.Fatalf("Error when creating server, reason: %v", err)
 	}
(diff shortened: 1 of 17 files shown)


authz/providers/azure/rbac/rbac.go
```diff
@@ -321,31 +364,33 @@ func (a *AccessInfo) sendCheckAccessRequest(checkAccessURL url.URL, checkAccessB
 		klog.V(10).Infof("binary data:%s", binaryData)
 	}
 
-	req, err := http.NewRequest(http.MethodPost, checkAccessURL.String(), buf)
+	req, err := http.NewRequestWithContext(ctx, http.MethodPost, checkAccessURL.String(), buf)
```
krdhruva, line 367: should we add `client request id` in each call and log that with error code?

authz/providers/azure/rbac/rbac.go
```diff
@@ -321,31 +364,33 @@ func (a *AccessInfo) sendCheckAccessRequest(checkAccessURL url.URL, checkAccessB
...
+		return errutils.WithCode(errors.Wrap(err, "error creating check access request"), http.StatusInternalServerError)
 	}
 
 	a.setReqHeaders(req)
-
+	internalServerCode := azureutils.ConvertIntToString(http.StatusInternalServerError)
+	// start time to calculate checkaccess duration
+	start := time.Now()
 	resp, err := a.client.Do(req)
+	duration := time.Since(start).Seconds()
```
krdhruva, line 377: Instead of emitting `checkAccessDuration` on multiple places, should we emit metric of duration just after call is returned?

authz/providers/azure/azure_test.go
```diff
@@ -167,4 +168,24 @@ func TestCheck(t *testing.T) {
...
+			User: "beta@bing.com",
+			ResourceAttributes: &authzv1.ResourceAttributes{
+				Namespace: "dev", Group: "", Resource: "pods",
+				Subresource: "status", Version: "v1", Name: "test", Verb: "delete",
+			}, Extra: map[string]authzv1.ExtraValue{"oid": {"00000000-0000-0000-0000-000000000000"}},
+		}
+
+		resp, err := client.Check(request, store)
+		assert.Nilf(t, resp, "response should be nil")
+		assert.NotNilf(t, err, "should get error")
```
krdhruva, line 188: nit: we can add validation for error code too.

authz/providers/azure/rbac/rbac.go
```diff
@@ -321,31 +364,33 @@ func (a *AccessInfo) sendCheckAccessRequest(checkAccessURL url.URL, checkAccessB
...
 	if err != nil {
-		reviewResult.err = errors.Wrap(err, "error in check access request execution")
-		ch <- reviewResult
-		return
+		checkAccessTotal.WithLabelValues(internalServerCode).Inc()
```
krdhruva, line 379: should we honor check access return code instead of returning internal server error?

authz/providers/azure/rbac/rbac.go
```diff
@@ -321,31 +364,33 @@ func (a *AccessInfo) sendCheckAccessRequest(checkAccessURL url.URL, checkAccessB
...
+		checkAccessDuration.WithLabelValues(internalServerCode).Observe(duration)
+		return errutils.WithCode(errors.Wrap(err, "error in check access request execution"), http.StatusInternalServerError)
```
krdhruva, line 381: Will gate way error & client time out be part of this error? If so, should we try to retrieve error code?

util/azure/utils.go
```diff
@@ -47,6 +50,45 @@ const (
...
+	discoverResourcesApiServerCallDuration = prometheus.NewHistogram(
+		prometheus.HistogramOpts{
+			Name:    "guard_apiresources_request_duration_seconds",
+			Help:    "A histogram of latencies for apiserver requests.",
+			Buckets: []float64{.25, .5, 1, 2.5, 5, 10, 15, 20},
+		})
+
+	discoverResourcesAzureCallDuration = prometheus.NewHistogram(
+		prometheus.HistogramOpts{
+			Name:    "guard_azure_get_operations_request_duration_seconds",
```
krdhruva, line 63: should we have request success/failure metric for this?

krdhruva (changes requested):
🕐

authz/providers/azure/rbac/rbac.go
Anumita: i know we had discussed this but i realized now that we can't do that -  we have code as dimension for checkaccess duration. so we can emit metric only after we know the code

authz/providers/azure/rbac/rbac.go
Anumita: resp and error cannot be non-nil at the same time

authz/providers/azure/azure_test.go
Anumita: added

authz/providers/azure/rbac/rbac.go
Anumita: adding

util/azure/utils.go
```diff
@@ -359,9 +416,12 @@ func fetchDataActionsList(settings *DiscoverResourcesSettings) ([]Operation, err
 	}
 
 	if resp.StatusCode != http.StatusOK {
+		counterGetOperationsResources.WithLabelValues(ConvertIntToString(resp.StatusCode)).Inc()
```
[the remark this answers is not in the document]
Anumita: as discussed , cannot move this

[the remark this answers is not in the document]
Anumita: also i removed this metric as we will never receive this metric in error case. Since, the endpoints of the guard service are assigned once the readiness probe has succeeded. So if it's not ready, prom will not be scrape the metrics anyway

authz/providers/azure/rbac/rbac.go
```diff
@@ -315,8 +318,16 @@ func (a *AccessInfo) CheckAccess(request *authzv1.SubjectAccessReviewSpec) (*aut
...
+			//create a request id for every checkaccess request
+			requestUUID := uuid.New()
+			reqContext := context.WithValue(egCtx, correlationRequestIDKey(correlationRequestIDHeader), []string{requestUUID.String()})
+			err := a.sendCheckAccessRequest(reqContext, checkAccessURL, body, ch)
 			if err != nil {
+				code := http.StatusInternalServerError
+				if v, ok := err.(errutils.HttpStatusCode); ok {
+					code = v.Code()
+				}
+				err = errutils.WithCode(errors.Errorf("Correlation ID: %s. Error: %s", requestUUID.String(), err), code)
```
krdhruva, line 332: nit: corr id should be after error.

krdhruva (dismissed):

util/error/utils.go
```diff
@@ -0,0 +1,68 @@
...
+
+func (w *withCode) Error() string { return w.cause.Error() }
+func (w *withCode) Cause() error  { return w.cause }
+func (w *withCode) Code() int     { return w.code }
+
+type HttpStatusCode interface {
+	Code() int
+}
+
+func (w *withCode) Format(s fmt.State, verb rune) {
```
weinong, line 51: where is rune defined?

authz/providers/azure/rbac/rbac.go
```diff
@@ -272,47 +308,69 @@ func (a *AccessInfo) CheckAccess(request *authzv1.SubjectAccessReviewSpec) (*aut
...
-		close(ch)
-	}()
-
-	var finalResult *authzv1.SubjectAccessReviewStatus
-	for result := range ch {
-		if result.err != nil {
-			return nil, result.err
+	if err := eg.Wait(); err != nil {
+		if ctx.Err() == context.DeadlineExceeded {
+			klog.V(5).Infof("Checkaccess requests have timed out. Error: %v\n", ctx.Err())
```
weinong, line 341: nit: no need to have \n

weinong (dismissed):
LGTM

util/error/utils.go
Anumita: rune is a type in golang - https://go.dev/doc/go1#rune

authz/providers/azure/rbac/rbac.go
Anumita: done

krdhruva (approved):

Anumita:
Hey @tamalsaha , could you please release image for guard? We have merged both the PR's Thanks!

Anumita:
Hey @tamalsaha , could you please release the image today if possible? We want to get this fix out asap, hence the urgency. Thanks!

PR #353, opened 2023-01-28
Add retries and timeouts to Azure auth provider "get metadata" step

AzureMarker:
We've seen issues where this "get metadata" step times out after 10s of minutes, by which time the HTTP server has already passed the write timeout and the API server returned Unauthorized since it didn't get a response from Guard.

This PR adds a timeout of 30 seconds and 3 retries to avoid failing due to transient network issues.

Also made a logging fix in the pop token code.


auth/providers/azure/azure.go
```diff
@@ -120,23 +123,30 @@ type metadataJSON struct {
...
 		return nil, err
 	}
-	defer response.Body.Close()
 
-	if response.StatusCode != http.StatusOK {
-		return nil, fmt.Errorf("get metadata url failed with status code: %d", response.StatusCode)
-	}
+	client := autorest.NewClientWithOptions(autorest.ClientOptions{})
+	client.Sender = httpclient.DefaultHTTPClient
+	client.PollingDuration = 30 * time.Second
```
weinong, line 133: isn't Get Metadata a sync call? what is this polling duration trying to address?

weinong:
what is the server write timeout and timeout from api-server? Is 30s a total timeout from 3 retries? or is it from each call?

auth/providers/azure/azure.go
AzureMarker: Polling duration is used as a per-request timeout:
https://pkg.go.dev/github.com/Azure/go-autorest/autorest#pkg-constants
https://pkg.go.dev/github.com/Azure/go-autorest/autorest#Client

AzureMarker:
> what is the server write timeout and timeout from api-server?

The server write timeout is coming from Guard, not API server. It's configurable, but set to 10 seconds by default:
https://github.com/kubeguard/guard/blob/09ab161fb1fb0862dba26c47a9b5c80c82a72b68/server/server.go#L54

> Is 30s a total timeout from 3 retries? or is it from each call?

It's the timeout per-retry. I could use `autorest.DoRetryForDuration` to have a total duration time limit.

auth/providers/azure/azure.go
weinong: i'm not convinced. https://github.com/search?q=repo%3AAzure%2Fgo-autorest%20PollingDuration&type=code
I don't see any code referencing this.

AzureMarker: Looks like you're right, I'm too used to working with async operations. Updated to include a timeout in the default HTTP client, but I can move that to only affect this code if requested.

amanohar (changes requested):
> Looks like you're right, I'm too used to working with async operations. Updated to include a timeout in the default HTTP client, but I can move that to only affect this code if requested. 

I am ok moving this to only retry for this specific use case.

AzureMarker:
> > Looks like you're right, I'm too used to working with async operations. Updated to include a timeout in the default HTTP client, but I can move that to only affect this code if requested. 
> 
> I am ok moving this to only retry for this specific use case.

The retries are only for this getMetadata operation, but the timeout change currently affects everything using the HTTP client.

weinong:
if it's 10s in server's write timeout, shouldn't the timeout be 3s for 3 times? besides, can we implement this short timeout just for that request?

AzureMarker:
Synced offline with Anubhuti - Changed timeout to only affect the getMetadata call.

> > if it's 10s in server's write timeout, shouldn't the timeout be 3s for 3 times? besides, can we implement this short timeout just for that request?
> 
> such as `http.NewRequestWithContext`

Reduced timeout to 3 seconds per attempt.
The issue with using `http.NewRequestWithContext` is that the timeout is implemented via deadline, so the request will instantly time out on the second attempt if the first attempt was a timeout (deadline didn't change since request is reused).
Setting the timeout on the HTTP client avoids this issue.

auth/providers/azure/azure.go
```diff
@@ -120,23 +122,34 @@ type metadataJSON struct {
...
 		return nil, err
 	}
-	defer response.Body.Close()
 
-	if response.StatusCode != http.StatusOK {
-		return nil, fmt.Errorf("get metadata url failed with status code: %d", response.StatusCode)
-	}
+	// Copy the default HTTP client so we can set a timeout.
+	// (It uses the same transport though since the pointer gets copied)
+	httpClient := *httpclient.DefaultHTTPClient
```
weinong, line 132: i don't think this is the right way. this will be creating new Transport and new connection pool every time

AzureMarker: It doesn't because the transport is a pointer / interface object, as the comment notes.

auth/providers/azure/azure.go
```diff
@@ -120,23 +122,34 @@ type metadataJSON struct {
 func getMetadata(aadEndpoint, tenantID string) (*metadataJSON, error) {
 	metadataURL := aadEndpoint + tenantID + "/.well-known/openid-configuration"
 
-	response, err := http.Get(metadataURL)
+	request, err := http.NewRequest("GET", metadataURL, nil)
```
bcho, line 125: If we are going to just set the timeout for the request, I wonder is that possible to just bind a context with 3s timeout to the request? The sender is just a normal http.Client, it should obey the timeout from request context

weinong: +100. NewRequestWithContext is designed for per request customization. this would be my preference as well.

weinong (approved):
LGTM

Anumita:
Hey @AzureMarker , we set the server write timeout to 25 seconds and server read timeout as 15 secs in aks

AzureMarker:
@tamalsaha can you publish a new release so we can start using this change? Thanks!

AzureMarker:
@tamalsaha may you publish a new release?

PR #354, opened 2023-02-10
Add support for reconciling discover resources

Anumita:
Discovering resources will be a best effort scenario. We may face errors while fetching the apiresources from apiserver as some apiservices may not be available. Also during upgrades,guard might call the old apiserver for the list. 

Added a reconciler which will updated the map every 5m. The frequency can be configured via flag.
It will not remove existing elements , only new elements will be added.

### DIFF
--- a/authz/providers/azure/options/options.go
+++ b/authz/providers/azure/options/options.go
@@ -19,6 +19,7 @@ package options
 import (
 	"fmt"
 	"strings"
+	"time"
 
 	"go.kubeguard.dev/guard/auth/providers/azure"
 
@@ -37,26 +38,28 @@ const (
 )
 
 type Options struct {
-	AuthzMode                       string
-	ResourceId                      string
-	AKSAuthzTokenURL                string
-	ARMCallLimit                    int
-	SkipAuthzCheck                  []string
-	SkipAuthzForNonAADUsers         bool
-	AllowNonResDiscoveryPathAccess  bool
-	UseNamespaceResourceScopeFormat bool
-	DiscoverResources               bool
-	KubeConfigFile                  string
+	AuthzMode                           string
+	ResourceId                          string
+	AKSAuthzTokenURL                    string
+	ARMCallLimit                        int
+	SkipAuthzCheck                      []string
+	SkipAuthzForNonAADUsers             bool
+	AllowNonResDiscoveryPathAccess      bool
+	UseNamespaceResourceScopeFormat     bool
+	DiscoverResources                   bool
+	ReconcileDiscoverResourcesFrequency time.Duration
+	KubeConfigFile                      string
 }
 
 func NewOptions() Options {
 	return Options{
-		ARMCallLimit:                    defaultArmCallLimit,
-		SkipAuthzCheck:                  []string{""},
-		SkipAuthzForNonAADUsers:         true,
-		AllowNonResDiscoveryPathAccess:  true,
-		UseNamespaceResourceScopeFormat: false,
-		DiscoverResources:               false,
+		ARMCallLimit:                        defaultArmCallLimit,
+		SkipAuthzCheck:                      []string{""},
+		SkipAuthzForNonAADUsers:             true,
+		AllowNonResDiscoveryPathAccess:      true,
+		UseNamespaceResourceScopeFormat:     false,
+		DiscoverResources:                   false,
+		ReconcileDiscoverResourcesFrequency: 10 * time.Minute,
 	}
 }
 
@@ -71,6 +74,7 @@ func (o *Options) AddFlags(fs *pflag.FlagSet) {
 	fs.BoolVar(&o.UseNamespaceResourceScopeFormat, "azure.use-ns-resource-scope-format", o.UseNamespaceResourceScopeFormat, "use namespace as resource scope format for making rbac checkaccess calls at namespace scope")
 	fs.StringVar(&o.KubeConfigFile, "azure.kubeconfig-file", "", "path to the kubeconfig of cluster.")
 	fs.BoolVar(&o.DiscoverResources, "azure.discover-resources", o.DiscoverResources, "fetch list of resources and operations from apiserver and azure. Default: false")
+	fs.DurationVar(&o.ReconcileDiscoverResourcesFrequency, "azure.discover-resources-frequency", o.ReconcileDiscoverResourcesFrequency, "Frequency at which discover resources should be reconciled. Default: 5m")
 }
 
 func (o *Options) Validate(azure azure.Options) []error {
--- a/authz/providers/azure/options/options.go
+++ b/authz/providers/azure/options/options.go
@@ -59,7 +59,7 @@ func NewOptions() Options {
 		AllowNonResDiscoveryPathAccess:      true,
 		UseNamespaceResourceScopeFormat:     false,
 		DiscoverResources:                   false,
-		ReconcileDiscoverResourcesFrequency: 10 * time.Minute,
+		ReconcileDiscoverResourcesFrequency: 5 * time.Minute,
 	}
 }
 
--- a/authz/providers/azure/rbac/checkaccessreqhelper.go
+++ b/authz/providers/azure/rbac/checkaccessreqhelper.go
@@ -47,7 +47,10 @@ const (
 	PodsResource                = "pods"
 )
 
-var username string
+var (
+	username               string
+	getStoredOperationsMap = azureutils.GetOperationsMap
+)
 
 type SubjectInfoAttributes struct {
 	ObjectId string   `json:"ObjectId"`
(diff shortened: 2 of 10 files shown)


util/azure/utils.go
```diff
@@ -48,6 +51,14 @@ const (
 	ConnectedClusters           = "Microsoft.Kubernetes/connectedClusters"
 	OperationsEndpointFormatARC = "%s/providers/Microsoft.Kubernetes/operations?api-version=2021-10-01"
 	OperationsEndpointFormatAKS = "%s/providers/Microsoft.ContainerService/operations?api-version=2018-10-31"
+	retrieveServerApiError      = "unable to retrieve the complete list of server APIs"
+	apiserviceUnreachableError  = "the server is currently unable to handle the request"
+)
+
+var (
+	OperationsMapLock   = sync.RWMutex{}
```
weinong, line 59: generally, it's not a good practice to have a globally accessible lock and its protected resource

util/azure/utils.go
```diff
@@ -162,39 +189,69 @@ func ConvertIntToString(number int) string {
...
+			err := DiscoverResources()
+			discoverResourcesDuration := time.Since(discoverResourcesListStart).Seconds()
+			if err != nil {
+				code := http.StatusInternalServerError
+				if v, ok := err.(errutils.HttpStatusCode); ok {
+					code = v.Code()
+				}
+				counterDiscoverResources.WithLabelValues(ConvertIntToString(code)).Inc()
+				DiscoverResourcesTotalDuration.Observe(discoverResourcesDuration)
+				klog.Errorf("Failed to reconcile map of data actions. Error:%s", err)
```
weinong, line 246: you should return here?

Anumita: since it's a reconciling loop, i thought we should not error out on error but instead just log it. hence not returning here

util/azure/utils.go
Anumita: done

authz/providers/azure/rbac/checkaccessreqhelper.go
```diff
@@ -240,11 +243,12 @@ func getDataActions(subRevReq *authzv1.SubjectAccessReviewSpec, clusterType stri
 			authInfoList = append(authInfoList, authInfoSingle)
 
 		} else {
-			if azureutils.GlobalOperationsMap == nil || (azureutils.GlobalOperationsMap != nil && len(azureutils.GlobalOperationsMap) == 0) {
+			storedOperationsMap := getStoredOperationsMap()
```
Anumita, line 246: check value or reference

util/azure/utils.go
```diff
@@ -153,6 +180,26 @@ func NewOperationsMap() OperationsMap {
...
+	operationsMapLock.Unlock()
+}
+
+func RLockOperationsMap() {
+	operationsMapLock.RLock()
+}
+
+func RUnlockOperationsMap() {
+	operationsMapLock.RUnlock()
+}
```
krdhruva, line 201: We are making its caller responsibility to maintain synchronization while accessing data. Instead let's explore if we can return copy with read lock to make read operation thread safe.

authz/providers/azure/options/options.go
```diff
@@ -68,6 +71,7 @@ func (o *Options) AddFlags(fs *pflag.FlagSet) {
 	fs.BoolVar(&o.UseNamespaceResourceScopeFormat, "azure.use-ns-resource-scope-format", o.UseNamespaceResourceScopeFormat, "use namespace as resource scope format for making rbac checkaccess calls at namespace scope")
 	fs.StringVar(&o.KubeConfigFile, "azure.kubeconfig-file", "", "path to the kubeconfig of cluster.")
 	fs.BoolVar(&o.DiscoverResources, "azure.discover-resources", o.DiscoverResources, "fetch list of resources and operations from apiserver and azure. Default: false")
+	fs.DurationVar(&o.ReconcileDiscoverResourcesFrequency, "azure.discover-resources-frequency", o.ReconcileDiscoverResourcesFrequency, "Frequency at which discover resources should be reconciled. Default: 5m")
```
krdhruva, line 74: this is not in sync with above value.

authz/providers/azure/rbac/checkaccessreqhelper.go
```diff
@@ -240,11 +243,12 @@ func getDataActions(subRevReq *authzv1.SubjectAccessReviewSpec, clusterType stri
 			authInfoList = append(authInfoList, authInfoSingle)
 
 		} else {
-			if operationsMap == nil || (operationsMap != nil && len(operationsMap) == 0) {
+			storedOperationsMap := getStoredOperationsMap()
```
krdhruva, line 246: should this have read lock before this call?

authz/providers/azure/options/options.go
Anumita: changed to 5m

util/azure/utils.go
Anumita: as discussed offline, instead of exposing the lock , in checkaccesshelper we will deepcopy the map and use that version for the calls

util/azure/utils.go
Anumita: on startup we are throwing any non apiservice error as fatal but during reconciling since this could be a transient error and we already would have a old version of the map that can be used we don't need to fatalf here but we can just show error here.

util/azure/utils.go
```diff
@@ -48,6 +51,14 @@ const (
 	ConnectedClusters           = "Microsoft.Kubernetes/connectedClusters"
 	OperationsEndpointFormatARC = "%s/providers/Microsoft.Kubernetes/operations?api-version=2021-10-01"
 	OperationsEndpointFormatAKS = "%s/providers/Microsoft.ContainerService/operations?api-version=2018-10-31"
+	retrieveServerApiError      = "unable to retrieve the complete list of server APIs"
+	apiserviceUnreachableError  = "the server is currently unable to handle the request"
+)
+
+var (
+	operationsMapLock = sync.RWMutex{}
+	operationsMap     = NewOperationsMap()
```
krdhruva, line 60: is this expected? I observed operationsMap being used in range, lenght functions..

util/azure/utils.go
```diff
@@ -162,39 +204,69 @@ func ConvertIntToString(number int) string {
...
+					code = v.Code()
+				}
+				counterDiscoverResources.WithLabelValues(ConvertIntToString(code)).Inc()
+				DiscoverResourcesTotalDuration.Observe(discoverResourcesDuration)
+				klog.Errorf("Failed to reconcile map of data actions. Error:%s", err)
+			}
 
-	return settings, nil
+			counterDiscoverResources.WithLabelValues(ConvertIntToString(http.StatusOK)).Inc()
+			DiscoverResourcesTotalDuration.Observe(discoverResourcesDuration)
```
krdhruva, line 265: should this be inside else?

util/azure/utils.go
```diff
@@ -204,37 +276,37 @@ func NewDiscoverResourcesSettings(clusterType string, environment string, loginU
...
-		return operationsMap, errors.Wrap(err, "Failed to fetch operations from Azure.")
+		return errors.Wrap(err, "Failed to fetch operations from Azure.")
 	}
 
 	discoverResourcesAzureCallDuration.Observe(getOperationsDuration)
 
-	operationsMap = createOperationsMap(apiResourcesList, operationsList, settings.clusterType)
+	createOperationsMap(apiResourcesList, operationsList)
 
 	klog.V(5).Infof("Operations Map created for resources: %s", operationsMap)
```
krdhruva, line 302: should this log line be moved to createOperationsMap?

util/azure/utils.go
```diff
@@ -338,20 +407,26 @@ func fetchApiResources(settings *DiscoverResourcesSettings) ([]*metav1.APIResour
...
 	if klog.V(5).Enabled() {
 		printApiresourcesList, _ := json.Marshal(apiresourcesList)
 
 		klog.Infof("List of ApiResources fetched from apiserver: %s", string(printApiresourcesList))
 	}
 
+	if err != nil {
+		// ignoring unreachable apiservice error.
+		if strings.Contains(err.Error(), retrieveServerApiError) && strings.Contains(err.Error(), apiserviceUnreachableError) {
+			klog.Infof("Error while fetching apiservices from apiserver: %s", err.Error())
```
krdhruva, line 420: It seems in this error case we are swallowing error and continuing with discover resource. Should we return err and skip further processing?

util/azure/utils.go
Anumita: no i think this is fine here since it's the final map so i want to print it after function is done

util/azure/utils.go
Anumita: don't understand , we need to skip the apiservice error , only if error is not the apiservice error then we need to return err as discussed

util/azure/utils.go
krdhruva: While printing, `operationsMap` is accessed without read lock. Hence question was to move it to func where its inside lock.

util/azure/utils.go
krdhruva: Can we add UT to verify partial list scenario to make sure existing data is not removed.

util/azure/utils.go
Anumita: it should be okay to do it without read lock since we are not modifying it at the same time, but i have added it in the other function just to be safe

util/azure/utils.go
Anumita: added test

PR #358, opened 2023-02-28
Use the request context throughout the code, and add retries & timeouts to more Azure auth provider calls

AzureMarker:
There are two improvements in this PR:
1. The request context is used. This context gets canceled when the client cancels the request or the connection closes, stopping operations from running longer than needed.
2. The Azure auth provider uses a retryable HTTP client with request timeouts wherever possible. This should improve the reliability of the provider when network connectivity is unstable.

Tested by deploying to an Azure Resource Bridge and verifying auth calls to Guard still worked.


auth/providers/azure/azure.go
```diff
@@ -139,8 +132,32 @@ func getMetadata(aadEndpoint, tenantID string) (*metadataJSON, error) {
...
+}
 
-	response, err := retryClient.Get(metadataURL)
+// withRetryableHttpClient sets the oauth2.HTTPClient key of the context to an
+// *http.Client made from makeRetryableHttpClient.
+// Some of the libraries we use will take the client out of the context via
+// oauth2.HTTPClient and use it, so this way we can add retries to external code.
+func withRetryableHttpClient(ctx context.Context) context.Context {
+	retryClient := makeRetryableHttpClient()
+	return context.WithValue(ctx, oauth2.HTTPClient, retryClient.StandardClient())
```
amanohar, line 143: I see this being used only in vendor/github.com/coreos/go-oidc/oidc.go. Are there other places where client is being read from the context to use the retries?

AzureMarker: That's the right place, though it's called in a few different ways (`oidc.NewProvider` and by `remoteKeySet.updateKeys`, which happens when we call `s.verifier.Verify(ctx, token)` in `Check`).

auth/providers/azure/azure.go
```diff
@@ -139,8 +132,32 @@ func getMetadata(aadEndpoint, tenantID string) (*metadataJSON, error) {
 		Backoff:      retryablehttp.DefaultBackoff,
```
Anumita, line 132: where are we setting the max no of retries here as 3?

AzureMarker: It's 3 attempts, not 3 retries.

```go
RetryMax: 2, // initial + 2 retries = 3 attempts
```

Anumita (approved):

Anumita (approved):
lgtm. @AzureMarker should we go ahead and merge?

AzureMarker:
> lgtm. @AzureMarker should we go ahead and merge?

Yes, if possible (we need an approval from someone who is authorized to merge).

krdhruva (approved):

PR #365, opened 2023-05-30
Add access token cnf public key validation in PoP token verifier

AzureMarker:
This check ensures that the user who requested the access token is the same user who created the PoP token.

Also added a few missed error cases and tests to the PoP token verifier.

### DIFF
--- a/auth/providers/azure/pop_tokenverifier.go
+++ b/auth/providers/azure/pop_tokenverifier.go
@@ -165,7 +165,7 @@ func (p *PoPTokenVerifier) ValidatePopToken(token string) (string, error) {
 	var at string
 	if atClaim, ok := claims["at"]; ok {
 		if at, ok = atClaim.(string); !ok {
-			return "", errors.Errorf("Invalid token. at claim should be string")
+			return "", errors.Errorf("Invalid token. 'at' claim should be string")
 		}
 	} else {
 		return "", errors.Errorf("Invalid token. access token missing")
--- a/auth/providers/azure/pop_tokenverifier.go
+++ b/auth/providers/azure/pop_tokenverifier.go
@@ -222,7 +222,7 @@ func (p *PoPTokenVerifier) verifyAccessTokenClaims(publicKey jwk, accessToken st
 		return fmt.Errorf("could not parse access token in PoP token: %w", err)
 	}
 
-	// Get claims
+	// Get claims (don't verify signature since that's done by a separate verifier later)
 	var claims Claims
 	err = accessTokenJwt.UnsafeClaimsWithoutVerification(&claims)
 	if err != nil {
--- a/auth/providers/azure/pop_tokenverifier.go
+++ b/auth/providers/azure/pop_tokenverifier.go
@@ -20,6 +20,7 @@ import (
 	"bytes"
 	"crypto"
 	"crypto/rsa"
+	"crypto/sha256"
 	"encoding/base64"
 	"encoding/binary"
 	"fmt"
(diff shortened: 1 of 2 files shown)


julienstroheker (approved):

auth/providers/azure/pop_tokenverifier.go
```diff
@@ -154,6 +157,23 @@ func (p *PoPTokenVerifier) ValidatePopToken(token string) (string, error) {
...
 
+	if jwk.N == "" || jwk.E == "" {
+		return "", errors.Errorf("Invalid token. 'jwk' claim is empty")
+	}
+
+	// Verify claims from access token
+	var at string
+	if atClaim, ok := claims["at"]; ok {
+		if at, ok = atClaim.(string); !ok {
+			return "", errors.Errorf("Invalid token. at claim should be string")
```
weinong, line 168:
```suggestion
			return "", errors.Errorf("Invalid token. 'at' claim should be string")
```

auth/providers/azure/pop_tokenverifier.go
```diff
@@ -193,6 +213,43 @@ func (p *PoPTokenVerifier) ValidatePopToken(token string) (string, error) {
...
+func (p *PoPTokenVerifier) verifyAccessTokenClaims(publicKey jwk, accessToken string) error {
+	// Parse JWT
+	accessTokenJwt, err := jwt.ParseSigned(accessToken)
+	if err != nil {
+		return fmt.Errorf("could not parse access token in PoP token: %w", err)
+	}
+
+	// Get claims
+	var claims Claims
+	err = accessTokenJwt.UnsafeClaimsWithoutVerification(&claims)
```
weinong, line 227: can you add a comment on why it should be without verification?

auth/providers/azure/pop_tokenverifier.go
```diff
@@ -193,6 +213,43 @@ func (p *PoPTokenVerifier) ValidatePopToken(token string) (string, error) {
...
+	if err != nil {
+		return fmt.Errorf("could not parse claims in access token: %w", err)
+	}
+
+	// Get CNF (confirmation) claim
+	cnf, ok := claims["cnf"]
+	if !ok {
+		return errors.New("could not retrieve cnf claim from access token")
```
weinong, line 235:
```suggestion
		return errors.New("could not retrieve 'cnf' claim from access token")
```

auth/providers/azure/pop_tokenverifier.go
```diff
@@ -193,6 +213,43 @@ func (p *PoPTokenVerifier) ValidatePopToken(token string) (string, error) {
...
+	}
+	cnfMap := map[string]string{}
+	if err := marshalGenericTo(cnf, &cnfMap); err != nil {
+		return fmt.Errorf("failed while parsing cnf in access token: %w", err)
```
weinong, line 239:
```suggestion
		return fmt.Errorf("failed while parsing 'cnf' in access token: %w", err)
```

auth/providers/azure/pop_tokenverifier.go
```diff
@@ -193,6 +213,43 @@ func (p *PoPTokenVerifier) ValidatePopToken(token string) (string, error) {
...
+	}
+
+	// Build the expected public key data
+	jwk := fmt.Sprintf(`{"e":"%s","kty":"RSA","n":"%s"}`, publicKey.E, publicKey.N)
+	jwkS256 := sha256.Sum256([]byte(jwk))
+	encodedJwt := base64.RawURLEncoding.EncodeToString(jwkS256[:])
+
+	if encodedJwt != cnfMap["kid"] {
+		return fmt.Errorf("PoP token validate failed: cnf claim mismatch")
```
weinong, line 248:
```suggestion
		return fmt.Errorf("PoP token validate failed: 'cnf' claim mismatch")
```

weinong:
i actually not quite follow how this change ensures that the user who requested the access token is the same user who created the PoP token

AzureMarker:
> i actually not quite follow how this change ensures that the user who requested the access token is the same user who created the PoP token

The user who created the inner access token gave the PoP token public key to AAD, and the returned access token contains the public key info in the `cnf` claim.

The PoP token wraps the access token, includes the same public key, and is signed by the PoP token private key. The PoP token signature can be verified using the included public key, and the public key in the PoP token claims can be checked against the inner access token's `cnf` claim.

auth/providers/azure/pop_tokenverifier.go
AzureMarker: Updated

weinong (dismissed):
LGTM

weinong (approved):
LGTM

weinong:
@@tamalsaha I plan to cut a release. Is there any instruction?

PR #367, opened 2023-06-14
Add nonce validation in PoP token verifier

julienstroheker:
Making sure nonce claim is not been re-used more than once on each requests.

- Validating if nonce claim is present as string
- Saving each `nonce` claim value into a map (cache)
- Validating that each requests is having a difference `nonce` claims by checking the values in the cache.
- Logging cache size via a log line.
- Cleaning the cache after TTL pop token expiration + 1minutes by spawning a go routine for each calls
> Since [each go routine will allocate `2kb` on the stack](https://cs.opensource.google/go/go/+/refs/tags/go1.20.5:src/runtime/stack.go;l=75), let's say if we do handle 100,000 calls this will cost 10Mb on the stack which I think is fair. In case of huge load, the TTL can be decreased.


auth/providers/azure/pop_tokenverifier.go
```diff
@@ -136,6 +143,28 @@ func (p *PoPTokenVerifier) ValidatePopToken(token string) (string, error) {
...
+		return "", errors.Errorf("Invalid token. 'nonce' claim is reused")
+	}
+	nonceMap[claims["nonce"].(string)] = time.Now()
+	nonceMapCounter++
+	klog.V(6).Infof("nonce claim added to the cache. Cache size is: %d", nonceMapCounter)
+	// Cleaning cached nonce token after PoPTokenValidityDuration minutes + 1 minute
+	go func() {
+		time.Sleep((p.PoPTokenValidityDuration + 1) * time.Minute)
+		delete(nonceMap, claims["nonce"].(string))
+		nonceMapCounter--
```
amanohar, line 165: Updating `nonceMapCounter` var should be made thread safe to prevent competing updates from multiple go routines.

amanohar (changes requested):

auth/providers/azure/pop_tokenverifier.go
```diff
@@ -34,6 +34,13 @@ import (
 	"k8s.io/klog/v2"
 )
 
+var (
+	// create a cache to save nonce claim to make sure the nonce is not reused
+	nonceMap = make(map[string]time.Time)
+	// counter to keep track of the number of nonce claims in the cache
+	nonceMapCounter = 0
```
AzureMarker, line 41: Can't we use `len(nonceMap)`?

auth/providers/azure/pop_tokenverifier.go
```diff
@@ -136,6 +143,28 @@ func (p *PoPTokenVerifier) ValidatePopToken(token string) (string, error) {
...
 	}
 
+	// Verify host 'nonce' claim
+	if nonce, ok := claims["nonce"]; ok {
+		if _, ok := nonce.(string); !ok {
+			return "", errors.Errorf("Invalid token. 'nonce' claim should be of string")
+		}
+	} else {
+		return "", errors.Errorf("Invalid token. 'nonce' claim is missing")
+	}
```
AzureMarker, line 153: Suggestion: Store nonce in a variable instead of indexing and casting for each usage.
```suggestion
	// Verify host 'nonce' claim
	var nonce string
	if nonceClaim, ok := claims["nonce"]; ok {
		if nonce, ok = nonceClaim.(string); !ok {
			return "", errors.Errorf("Invalid token. 'nonce' claim should be of string")
		}
	} else {
		return "", errors.Errorf("Invalid token. 'nonce' claim is missing")
	}
```

auth/providers/azure/pop_tokenverifier.go
```diff
@@ -136,6 +143,28 @@ func (p *PoPTokenVerifier) ValidatePopToken(token string) (string, error) {
...
+	// Making sure nonce is not reused
+	if _, ok := nonceMap[claims["nonce"].(string)]; ok {
```
AzureMarker, line 155: Maps aren't thread safe, we need to have all access behind a lock:
https://stackoverflow.com/questions/36167200/how-safe-are-golang-maps-for-concurrent-read-write-operations

auth/providers/azure/pop_tokenverifier_test.go
```diff
@@ -239,63 +241,68 @@ func GeneratePoPToken(ts int64, hostName, kid string) (string, error) {
...
 	verifier := NewPoPVerifier("testHostname", 15*time.Minute)
 
-	validToken, _ := GeneratePoPToken(time.Now().Unix(), "testHostname", "")
+	validToken, _ := GeneratePoPToken(time.Now().Unix(), "testHostname", "", "randomnonce")
 	_, err := verifier.ValidatePopToken(validToken)
 	assert.NoError(t, err)
 
-	invalidToken, _ := GeneratePoPToken(time.Now().Unix(), "", badTokenKey)
+	// Test for replay attack
+	validToken, _ = GeneratePoPToken(time.Now().Unix(), "testHostname", "", "randomnonce")
```
AzureMarker, line 249: For a replay attack, don't regenerate the token. There can be a separate test which uses a different timestamp but same nonce.

AzureMarker (changes requested):

auth/providers/azure/pop_tokenverifier.go
```diff
@@ -27,23 +27,66 @@ import (
...
 )
 
+// create a cache to save nonce claim to make sure the nonce is not reused
+var nonceMap = nonceCache{v: make(map[string]time.Time)}
+
+type nonceCache struct {
+	mu sync.Mutex
+	v  map[string]time.Time
+	// counter to keep track of the number of nonce claims in the cache
+	counter int
```
AzureMarker, line 45: Why do we need this counter when we can do `len(v)`?

auth/providers/azure/pop_tokenverifier.go
```diff
@@ -27,23 +27,66 @@ import (
...
+}
+
+func (c *nonceCache) Dec(key string) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	c.counter--
+}
```
AzureMarker, line 52: This function isn't used (and makes the counter inconsistent).

julienstroheker: yep I did clean it

auth/providers/azure/pop_tokenverifier.go
julienstroheker: Sure

auth/providers/azure/pop_tokenverifier_test.go
```diff
@@ -237,65 +240,140 @@ func GeneratePoPToken(ts int64, hostName, kid string) (string, error) {
...
+	verifierNonce.mapCacheRetentionBuffer = 0
+	// Testing map cache cleanup
+	validToken, _ = GeneratePoPToken(time.Now().Unix(), "testHostname", "", "randomnonce1")
+	_, err = verifierNonce.ValidatePopToken(validToken)
+	assert.NoError(t, err)
+	// Wait for cache to be cleaned up
+	time.Sleep(2 * time.Second)
+
+	_, err = verifierNonce.ValidatePopToken(validToken)
+	assert.NoError(t, err)
```
AzureMarker, line 323: Shouldn't this have an error since the PoP token is only valid for 1 second according to the `NewPoPVerifier` call on line 313, and it's been 2 seconds since the token was created?

auth/providers/azure/pop_tokenverifier_test.go
```diff
@@ -237,65 +240,140 @@ func GeneratePoPToken(ts int64, hostName, kid string) (string, error) {
...
+			time.Sleep(1 * time.Second)
+		}
+		return tokens[0], tokens[1], nil
+	}
+	token1, token2, err := getTokens(time.Now().Unix())
+	assert.NoError(t, err)
+
+	verifier := NewPoPVerifier(hostnameClaim, 1*time.Second)
+	_, err = verifier.ValidatePopToken(token1)
+	assert.NoError(t, err)
```
AzureMarker, line 375: This should fail because the token is only supposed to be valid for 1 second and it's already been at least 1 second since it was generated. Try changing the token TTL in the call to `NewPoPVerifier` (but also make sure the current test fails).

auth/providers/azure/pop_tokenverifier_test.go
```diff
@@ -237,65 +240,140 @@ func GeneratePoPToken(ts int64, hostName, kid string) (string, error) {
...
+				return "", "", fmt.Errorf("Error while signing pop key. Error:%+v", err)
+			}
+			signatureB64 := base64.RawURLEncoding.EncodeToString(signature)
+			tokens = append(tokens, headerB64+"."+payloadB64+"."+signatureB64)
+			// enforcing different timestamp
+			time.Sleep(1 * time.Second)
+		}
+		return tokens[0], tokens[1], nil
+	}
+	token1, token2, err := getTokens(time.Now().Unix())
```
AzureMarker, line 370: Why can't this be two calls to `GeneratePoPToken` with different timestamps and same nonce? `now-1` and `now`?

julienstroheker: Token will not be the same, what you describe is what I already do in the previous test

AzureMarker: I mean that we can reuse the `GeneratePoPToken` function instead of making this `getToken` closure.

julienstroheker: I was avoiding refactoring the whole code here but sure if you want I can take a look.

auth/providers/azure/pop_tokenverifier_test.go
julienstroheker: actually I think there is a bug in current implementation. Taking a look.

auth/providers/azure/pop_tokenverifier_test.go
AzureMarker: This works, though the difference is that the keys are not the same between the two tokens:
```go
token1, err := GeneratePoPToken(time.Now().Add(-1*time.Second).Unix(), hostnameClaim, "", nonceClaim)
assert.NoError(t, err)
token2, err := GeneratePoPToken(time.Now().Unix(), hostnameClaim, "", nonceClaim)
assert.NoError(t, err)
```

weinong:
@julienstroheker is this PR still open?

julienstroheker:
Yes. I was waiting the UT refactor pr to merge first.

auth/providers/azure/pop_tokenverifier.go
```diff
@@ -136,6 +169,28 @@ func (p *PoPTokenVerifier) ValidatePopToken(token string) (string, error) {
 		return "", errors.Errorf("Invalid token. 'u' claim is missing")
 	}
 
+	// Verify host 'nonce' claim
+	var nonce string
+	if nonceClaim, ok := claims["nonce"]; ok {
```
AzureMarker, line 174: I think these checks should be at the end of the function, so we only check and update the nonce cache if we've confirmed the caller has signed the token correctly (and it's not just someone else changing the token without resigning).

auth/providers/azure/pop_tokenverifier.go
```diff
@@ -136,6 +169,28 @@ func (p *PoPTokenVerifier) ValidatePopToken(token string) (string, error) {
...
+	}
+	// Making sure nonce is not reused
+	err = nonceMap.AddToCache(nonce)
+	if err != nil {
+		return "", errors.Errorf("Invalid token. 'nonce' claim is reused")
+	}
+	klog.V(6).Infof("nonce claim added to the cache. Cache size is: %d", nonceMap.GetCounter())
+	// Cleaning cached nonce token after PoPTokenValidityDuration minutes + 1 minute by default
+	go func() {
+		time.Sleep((p.PoPTokenValidityDuration + p.mapCacheRetentionBuffer))
```
AzureMarker, line 190: nit: extra parentheses aren't needed

auth/providers/azure/pop_tokenverifier.go
```diff
@@ -136,6 +169,28 @@ func (p *PoPTokenVerifier) ValidatePopToken(token string) (string, error) {
 		return "", errors.Errorf("Invalid token. 'u' claim is missing")
 	}
 
+	// Verify host 'nonce' claim
+	var nonce string
+	if nonceClaim, ok := claims["nonce"]; ok {
+		if _, ok := nonceClaim.(string); !ok {
+			return "", errors.Errorf("Invalid token. 'nonce' claim should be of string")
```
AzureMarker, line 176: Typo: `'nonce' claim should be of type string`

auth/providers/azure/pop_tokenverifier_test.go
```diff
@@ -219,4 +231,13 @@ func TestPopTokenVerifier_Verify(t *testing.T) {
 		_, err := verifier.ValidatePopToken(invalidToken)
 		assert.Containsf(t, err.Error(), "Token is expired", "Error message is not as expected")
 	})
+
+	t.Run("'nonce' claim been reused.", func(t *testing.T) {
```
AzureMarker, line 235: Add a test to verify that you can reuse a nonce claim once the cache entry expires.

auth/providers/azure/pop_tokenverifier_test.go
```diff
@@ -219,4 +231,17 @@ func TestPopTokenVerifier_Verify(t *testing.T) {
...
+	t.Run("'nonce' claim been reused - Cache validation", func(t *testing.T) {
+		nonceVerifier := azure.NewPoPVerifier("testHostname", 4*time.Second, -3*time.Second)
+		validToken, _ := azure.NewPoPTokenBuilder().SetTimestamp(time.Now().Unix()).SetHostName("testHostname").SetKid(azure.NonceClaimHardcoded).GetToken()
+		_, err := nonceVerifier.ValidatePopToken(validToken)
+		assert.NoError(t, err)
+		validToken, _ = azure.NewPoPTokenBuilder().SetTimestamp(time.Now().Unix()).SetHostName("testHostname").SetKid(azure.NonceClaimHardcoded).GetToken()
+		_, err = nonceVerifier.ValidatePopToken(validToken)
+		assert.Containsf(t, err.Error(), "Invalid token. 'nonce' claim is reused", "Error message is not as expected")
+		time.Sleep(2 * time.Second)
+		_, err = nonceVerifier.ValidatePopToken(validToken)
```
AzureMarker, line 244: nit: Add comments explaining each test (first usage is fresh nonce, second is reused so error, third is reused after expired so allowed).

AzureMarker (approved):

AzureMarker (approved):

auth/providers/azure/pop_tokenverifier.go
```diff
@@ -27,23 +27,56 @@ import (
...
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	return len(c.v)
+}
+
 // PopTokenVerifier is validator for PoP tokens.
 type PoPTokenVerifier struct {
 	hostName                 string
 	PoPTokenValidityDuration time.Duration
+	mapCacheRetentionBuffer  time.Duration
```
weinong, line 72:
```suggestion
	mapCacheRetentionDuration  time.Duration
```

auth/providers/azure/pop_tokenverifier.go
```diff
@@ -210,6 +243,28 @@ func (p *PoPTokenVerifier) ValidatePopToken(token string) (string, error) {
...
+		return "", errors.Errorf("Invalid token. 'nonce' claim is missing")
+	}
+	// Making sure nonce is not reused
+	err = nonceMap.AddToCache(nonce)
+	if err != nil {
+		return "", errors.Errorf("Invalid token. 'nonce' claim is reused")
+	}
+	klog.V(6).Infof("nonce claim added to the cache. Cache size is: %d", nonceMap.GetCounter())
+	// Cleaning cached nonce token after PoPTokenValidityDuration minutes + 1 minute by default
+	go func() {
```
weinong, line 263: there is already a similar implementation that is uses a more efficient library. can we just do the same?
https://github.com/kubeguard/guard/blob/master/authz/providers/azure/data/datastore.go

weinong: for instance, this implementation could probably use some jitter that's already implemented in those libraries

julienstroheker:
Still holding this PR due to changes we do on RBAC. This requires more tests + to review PR comments.

# PLEASE DO NOT MERGE

auth/providers/azure/pop_tokenverifier.go
julienstroheker: Refactored to make it re-usable and leveraging it

auth/providers/azure/pop_tokenverifier.go
```diff
@@ -29,22 +29,71 @@ import (
...
 	"time"
 
+	cache "go.kubeguard.dev/guard/util/data"
+
 	"github.com/pkg/errors"
 	"gopkg.in/square/go-jose.v2/jwt"
 	"k8s.io/klog/v2"
 )
 
+// // create a cache to save nonce claim to make sure the nonce is not reused
```
julienstroheker, line 39: this can be removed

PR #368, opened 2023-06-21
Fix expiration pop token duration option

julienstroheker:
Currently conversion `PoPTokenValidityDuration` is not acting as expected. 

When passing a duration such as `15m` current code was multiplying `15m` (900000000000) with `time.Minute` (60000000000) with was causing very high expiration date non expected.

This fix assumes options are set using a proper time.Duration and respect it down to the pop token verification with any conversion in between.

### DIFF
--- a/auth/providers/azure/options.go
+++ b/auth/providers/azure/options.go
@@ -74,7 +74,7 @@ func (o *Options) AddFlags(fs *pflag.FlagSet) {
 	fs.StringVar(&o.AKSTokenURL, "azure.aks-token-url", "", "url to call for AKS OBO flow")
 	fs.StringVar(&o.POPTokenHostname, "azure.pop-hostname", "", "hostname used to run the pop hostname verification; 'u' claim")
 	fs.BoolVar(&o.EnablePOP, "azure.enable-pop", false, "Enabling pop token verification")
-	fs.DurationVar(&o.PoPTokenValidityDuration, "azure.pop-token-validity-duration", 15, "time duration for PoP token to be considered valid from creation time, default 15 min")
+	fs.DurationVar(&o.PoPTokenValidityDuration, "azure.pop-token-validity-duration", 15*time.Minute, "time duration for PoP token to be considered valid from creation time, default 15 min")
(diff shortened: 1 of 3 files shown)


auth/providers/azure/options.go
```diff
@@ -74,7 +74,7 @@ func (o *Options) AddFlags(fs *pflag.FlagSet) {
 	fs.StringVar(&o.AKSTokenURL, "azure.aks-token-url", "", "url to call for AKS OBO flow")
 	fs.StringVar(&o.POPTokenHostname, "azure.pop-hostname", "", "hostname used to run the pop hostname verification; 'u' claim")
 	fs.BoolVar(&o.EnablePOP, "azure.enable-pop", false, "Enabling pop token verification")
-	fs.DurationVar(&o.PoPTokenValidityDuration, "azure.pop-token-validity-duration", 15, "time duration for PoP token to be considered valid from creation time, default 15 min")
+	fs.DurationVar(&o.PoPTokenValidityDuration, "azure.pop-token-validity-duration", 15*time.Minute, "time duration for PoP token to be considered valid from creation time, default 15 min")
```
AzureMarker, line 77: Passing in nanoseconds (`time.Duration`'s underlying unit) seems unexpected for a CLI argument. Maybe we can change the type here to be `int` and rename the variable to `PoPTokenValidityMinutes`? Then do the conversion to `time.Duration` later.

julienstroheker: `Passing in nanoseconds (time.Duration's underlying unit) seems unexpected for a CLI argument` why is that ? there is a `fs.DurationVar` we should just use it. This also allows to pass seconds / minutes / hours if needed.

AzureMarker: nvm, it seems that Go is able to parse `15m` into a duration so users can pass it via a human-readable format.

AzureMarker (approved):

amanohar (approved):

weinong (approved):
LGTM

AzureMarker:
@tamalsaha Can you make a new release for this?

tamalsaha:
https://github.com/kubeguard/guard/releases/tag/v0.15.1

PR #369, opened 2023-06-28
Update RBAC API version

dvob:
Update RABC API version because rbac.authorization.k8s.io/v1beta1 got removed in Kubernetes 1.22 (https://kubernetes.io/docs/reference/using-api/deprecation-guide/#rbac-resources-v122).

Another question I asked my self is why this cluster role to list nodes is required at all? could possible removed entirely?


tamalsaha (approved):

tamalsaha:
Thanks, @dvob !

> Another question I asked my self is why this cluster role to list nodes is required at all? could possible removed entirely?

I don't think it is needed anymore. It should be ok to remove.

dvob:
@tamalsaha ok, I will update the PR accordingly to remove these two resources.

dvob:
@tamalsaha I removed the cluster role, cluster role binding and the service account which were required to list the nodes.
then I also removed the annotation "scheduler.alpha.kubernetes.io/critical-pod" which is no longer used in kubernetes clusters 1.16+

tamalsaha (approved):

PR #371, opened 2023-07-31
Refactor and improve UT for PoP token verifier 

nehaPrakash98:
* The goal of this PR is to improve the test coverage. All of the happy path and error cases (where feasible) are covered by unit tests.
* In order to improve the readability of the tests, the test code that makes the test PoP tokens has been refactored. The existing code was hard to extend for new scenarios and hard to follow. The new code uses a builder pattern to create the tokens.

Previous test coverage was ~80%
Current test coverage is 91.5%

 
image

### DIFF
--- a/auth/providers/azure/pop_tokenverifier_test.go
+++ b/auth/providers/azure/pop_tokenverifier_test.go
@@ -14,460 +14,209 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package azure
+package azure_test
 
 import (
-	"crypto"
-	"crypto/rand"
-	"crypto/rsa"
-	"crypto/sha256"
-	"encoding/base64"
 	"fmt"
-	"math/big"
-	"strconv"
-	"strings"
 	"testing"
 	"time"
 
-	"github.com/google/uuid"
+	"go.kubeguard.dev/guard/auth/providers/azure"
+
 	"github.com/stretchr/testify/assert"
-	"gopkg.in/square/go-jose.v2"
 )
-
-const (
-	popAccessToken           = `{ "aud": "client", "iss" : "kd", "exp" : "%d","cnf": {"kid":"%s","xms_ksl":"sw"} }`
-	popAccessTokenWithoutCnf = `{ "aud": "client", "iss" : "kd", "exp" : "%d" }`
-)
-
-type swPoPKey struct {
-	key    *rsa.PrivateKey
-	keyID  string
-	jwk    string
-	jwkTP  string
-	reqCnf string
-}
-
-func (swk *swPoPKey) Alg() string {
-	return "RS256"
-}
-
-func (swk *swPoPKey) Sign(payload []byte) ([]byte, error) {
-	return swk.key.Sign(rand.Reader, payload, crypto.SHA256)
-}
-
-func (swk *swPoPKey) KeyID() string {
-	return swk.keyID
-}
-
-func (swk *swPoPKey) Cnf() string {
-	return swk.reqCnf
-}
-
-func (swk *swPoPKey) Jwk() string {
-	return swk.jwk
-}
-
-func NewSWPoPKey() (*swPoPKey, error) {
-	pop := &swPoPKey{}
-	rsa, err := rsa.GenerateKey(rand.Reader, 1028)
-	if err != nil {
-		return nil, err
-	}
-	pop.key = rsa
-	pubKey := rsa.PublicKey
-	e := big.NewInt(int64(pubKey.E))
-	eB64 := base64.RawURLEncoding.EncodeToString(e.Bytes())
-	n := pubKey.N
-	nB64 := base64.RawURLEncoding.EncodeToString(n.Bytes())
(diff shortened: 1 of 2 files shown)


azure-coverage.out
julienstroheker, line 1: We don't need this file as long with the html one right ?

nehaPrakash98: Oh right, I'll remove them

auth/providers/azure/pop_tokenverifier_test.go
```diff
@@ -243,6 +268,10 @@ func TestPopTokenVerifier_Verify(t *testing.T) {
 	_, err := verifier.ValidatePopToken(validToken)
 	assert.NoError(t, err)
 
+	validToken, _ = GeneratePoPToken(time.Now().Unix(), "testHostname", tsClaimsTypeString)
```
julienstroheker, line 271: remind me what is that test compared to the previous one ?

Also no harm adding a description comment line on top of each test.

AzureMarker:
Make sure to fix the DCO check:
https://github.com/kubeguard/guard/pull/371/checks?check_run_id=15500422959

auth/providers/azure/pop_tokenverifier_test.go
nehaPrakash98: Added comments for all tests

auth/providers/azure/pop_tokenverifier_test.go
```diff
@@ -143,160 +144,330 @@ const (
...
-	var header, headerB64 string
-	nonce := uuid.New().String()
-	nonce = strings.Replace(nonce, "-", "", -1)
-	keyID := popKey.KeyID()
-	algo := popKey.Alg()
-	typ := "pop"
-	if kid == headerBadKeyID {
-		keyID = ""
+
+	payload := fmt.Sprintf(`{ "at" : "%s", "ts" : %d, "u": "%s", "cnf":{"jwk":%s}, "nonce":"%s"}`, at, b.ts, b.hostName, b.popKey.Jwk(), nonce)
```
julienstroheker, line 252: Can this be a struct as well ?

auth/providers/azure/pop_tokenverifier_test.go
```diff
@@ -24,6 +24,7 @@ import (
 	"encoding/base64"
```
julienstroheker, line 24: Can the package be converted to be a `package azure_test` ?

auth/providers/azure/pop_tokenverifier_test.go
```diff
@@ -143,160 +144,330 @@ const (
...
+	tsClaimsTypeUnknown = "tsClaimsTypeUnknown"
+	uClaimsWrongType    = "uClaimsWrongType"
+	signatureWrongType  = "signatureWrongType"
 )
 
-func GeneratePoPToken(ts int64, hostName, kid string) (string, error) {
-	if strings.Contains(kid, badTokenKey) {
-		return kid, nil
+// A struct that represents a PoP token
+type PoPToken struct {
```
julienstroheker, line 164: Can we move all these helper struct / function into a dedicated file (`pop_tokenverifier_utils_test.go` ?) and just have the `TestPopTokenVerifier_Verify(t *testing.T)` func in this one ?

auth/providers/azure/pop_tokenverifier_test.go
```diff
@@ -143,160 +144,330 @@ const (
...
-	finalToken := headerB64 + "." + payloadB64 + "." + signatureB64
+	finalToken := b.token.Header + "." + b.token.Payload + "." + b.token.Signature
 	return finalToken, nil
 }
 
 func TestPopTokenVerifier_Verify(t *testing.T) {
 	verifier := NewPoPVerifier("testHostname", 15*time.Minute)
 
-	validToken, _ := GeneratePoPToken(time.Now().Unix(), "testHostname", "")
+	validToken, _ := NewPoPTokenBuilder().SetTimestamp(time.Now().Unix()).SetHostName("testHostname").GetToken()
```
julienstroheker, line 344: Can this be refactored to use a TDT (Table driven test) approach ?

```
func Test(t *testing.T) {
	testCases := []struct {
		desc	string
		
	}{
		{
			desc: "",
			
		},
	}
	for _, tC := range testCases {
		t.Run(tC.desc, func(t *testing.T) {
			
		})
	}
}
```

julienstroheker (changes requested):
Overall LGTM, this is really good improvment.

I added comment about structure.

auth/providers/azure/pop_tokenverifier_test.go
nehaPrakash98: If we change this to a struct, then type of each of the fields will be fixed. But, some of the test cases are to identify errors when the wrong type of argument is passed. If we make it a struct then we won't be able to test those. For example: the right type of `ts` is int64, but in `TsClaimsTypeString` case, we pass `ts` as a string.

julienstroheker:
@weinong are you able to trigger BUILD to validates the UT on this PR please ?

julienstroheker:
@nehaPrakash98 please check the DCO

julienstroheker (approved):
LGTM

weinong (approved):
LGTM

PR #372, opened 2023-08-07
Use retryable http client in Azure authz provider

bingosummer:
We see the timeout error in the checkaccess requrests sometimes.

```
Error occured during authorization check. Please retry again. Error: Checkaccess requests have timed out.: context deadline exceeded
```

This PR moves `withRetryableHttpClient` from Azure authn provider to `utils`, and leverage it in Azure authz provider's `RefreshToken` and `CheckAccess`.

### DIFF
--- a/authz/providers/azure/azure.go
+++ b/authz/providers/azure/azure.go
@@ -26,6 +26,7 @@ import (
 	"go.kubeguard.dev/guard/authz"
 	authzOpts "go.kubeguard.dev/guard/authz/providers/azure/options"
 	"go.kubeguard.dev/guard/authz/providers/azure/rbac"
+	azureutils "go.kubeguard.dev/guard/util/azure"
 	errutils "go.kubeguard.dev/guard/util/error"
 
 	"github.com/Azure/go-autorest/autorest/azure"
@@ -49,7 +50,8 @@ func init() {
 }
 
 type Authorizer struct {
-	rbacClient *rbac.AccessInfo
+	rbacClient           *rbac.AccessInfo
+	httpClientRetryCount int
 }
 
 func New(opts authzOpts.Options, authopts auth.Options) (authz.Interface, error) {
@@ -64,7 +66,9 @@ func New(opts authzOpts.Options, authopts auth.Options) (authz.Interface, error)
 }
 
 func newAuthzClient(opts authzOpts.Options, authopts auth.Options) (authz.Interface, error) {
-	c := &Authorizer{}
+	c := &Authorizer{
+		httpClientRetryCount: authopts.HttpClientRetryCount,
+	}
 
 	authzInfoVal, err := getAuthzInfo(authopts.Environment)
 	if err != nil {
@@ -119,6 +123,8 @@ func (s Authorizer) Check(ctx context.Context, request *authzv1.SubjectAccessRev
 		_ = s.rbacClient.SetResultInCache(request, true, store)
 		return &authzv1.SubjectAccessReviewStatus{Allowed: true, Reason: rbac.AccessAllowedVerdict}, nil
 	}
+
+	ctx = azureutils.WithRetryableHttpClient(ctx, s.httpClientRetryCount)
 
 	if s.rbacClient.IsTokenExpired() {
 		if err := s.rbacClient.RefreshToken(ctx); err != nil {
--- a/authz/providers/azure/azure_test.go
+++ b/authz/providers/azure/azure_test.go
@@ -18,6 +18,7 @@ package azure
 
 import (
 	"context"
+	"fmt"
 	"net"
 	"net/http"
 	"net/http/httptest"
@@ -37,7 +38,8 @@ import (
 )
 
 const (
-	loginResp = `{ "token_type": "Bearer", "expires_in": 8459, "access_token": "%v"}`
+	loginResp            = `{ "token_type": "Bearer", "expires_in": 8459, "access_token": "%v"}`
+	httpClientRetryCount = 2
 )
 
 func clientSetup(serverUrl, mode string) (*Authorizer, error) {
@@ -52,9 +54,10 @@ func clientSetup(serverUrl, mode string) (*Authorizer, error) {
 	}
 
 	authOpts := auth.Options{
-		ClientID:     "client_id",
-		ClientSecret: "client_secret",
-		TenantID:     "tenant_id",
+		ClientID:             "client_id",
+		ClientSecret:         "client_secret",
+		TenantID:             "tenant_id",
+		HttpClientRetryCount: httpClientRetryCount,
 	}
 
 	authzInfo := rbac.AuthzInfo{
@@ -171,6 +174,7 @@ func TestCheck(t *testing.T) {
 		assert.Nilf(t, resp, "response should be nil")
 		assert.NotNilf(t, err, "should get error")
 		assert.Contains(t, err.Error(), "Error occured during authorization check")
+		assert.Contains(t, err.Error(), fmt.Sprintf("giving up after %d attempt", httpClientRetryCount+1))
 		if v, ok := err.(errutils.HttpStatusCode); ok {
 			assert.Equal(t, v.Code(), http.StatusInternalServerError)
 		}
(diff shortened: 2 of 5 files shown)


karataliu (approved):

bingosummer:
@weinong @AzureMarker could you please have a review? Thanks!

authz/providers/azure/azure.go
```diff
@@ -49,7 +50,8 @@ func init() {
 }
 
 type Authorizer struct {
-	rbacClient *rbac.AccessInfo
+	rbacClient           *rbac.AccessInfo
+	HttpClientRetryCount int
```
weinong, line 54: should this property be exported (public)?

authz/providers/azure/azure.go
```diff
@@ -120,6 +124,8 @@ func (s Authorizer) Check(ctx context.Context, request *authzv1.SubjectAccessRev
 		return &authzv1.SubjectAccessReviewStatus{Allowed: true, Reason: rbac.AccessAllowedVerdict}, nil
 	}
 
+	ctx = azureutils.WithRetryableHttpClient(ctx, s.HttpClientRetryCount)
```
weinong, line 127: 1. why do we save the http client in the context?
2. as far as I can tell, it still uses [AccessInfo.client](https://github.com/kubeguard/guard/blob/master/authz/providers/azure/rbac/rbac.go#L160), no?

bingosummer: I followed the comment of `WithRetryableHttpClient`.   @AzureMarker **Some of the libraries** respects the http client in the context. Does it mean that it's a best-effort?

```
// WithRetryableHttpClient sets the oauth2.HTTPClient key of the context to an
// *http.Client made from makeRetryableHttpClient.
// Some of the libraries we use will take the client out of the context via
// oauth2.HTTPClient and use it, so this way we can add retries to external code.
```

authz/providers/azure/azure.go
bingosummer: fixed

authz/providers/azure/azure.go
AzureMarker: `net/http` is a standard library, it doesn't reference third party code so there's no way it uses that context key.

This key is used in go-oidc:
https://github.com/kubeguard/guard/blob/e0840f9ce43103cd006c07b1cdb4373125956d3e/vendor/github.com/coreos/go-oidc/oidc.go#L60

The other HTTP client usages hard-code to `go.kubeguard.dev/guard/util/httpclient`'s `DefaultHTTPClient`, which is just a normal HTTP client.

bingosummer: I made this change https://github.com/kubeguard/guard/pull/372/commits/57e588e83569f709f8fd04b22b8dadca66c24080. Could you please review it?

bingosummer: @AzureMarker gentle ping

auth/providers/azure/graph/obo_tokenprovider.go
```diff
@@ -74,6 +75,9 @@ func (u *oboTokenProvider) Acquire(ctx context.Context, token string) (AuthRespo
 		klog.V(10).Infoln(cmd)
 	}
 
+	if c, ok := ctx.Value(oauth2.HTTPClient).(*http.Client); ok {
+		u.client = c
```
AzureMarker, line 79: This shouldn't modify the token provider (`u.client`). Store a reference to the HTTP client in a variable and modify it there.

The issue is if this is called twice (first with a context that holds an HTTP client, then with a context that has no HTTP client), the second call will use the HTTP client of the first.

authz/providers/azure/rbac/rbac.go
```diff
@@ -374,6 +375,10 @@ func (a *AccessInfo) CheckAccess(request *authzv1.SubjectAccessReviewSpec) (*aut
 }
 
 func (a *AccessInfo) sendCheckAccessRequest(ctx context.Context, checkAccessURL url.URL, checkAccessBody *CheckAccessRequest, ch chan *authzv1.SubjectAccessReviewStatus) error {
+	if c, ok := ctx.Value(oauth2.HTTPClient).(*http.Client); ok {
+		a.client = c
```
AzureMarker, line 379: Same comment here, don't modify `a.client`, store it in a local variable.

authz/providers/azure/azure.go
AzureMarker: Added some comments

auth/providers/azure/graph/obo_tokenprovider.go
bingosummer: removed this part to make the pr focus on authz.

authz/providers/azure/rbac/rbac.go
bingosummer: fixed. Could you please check whether the fix meets the requirement? Thanks

bingosummer:
@AzureMarker  Could you please have a review again? Thanks

AzureMarker (approved):

bingosummer:
@weinong could you please have a review? Thanks

bingosummer:
The PR is rebased. @weinong could you please have a review? Please let me know if anything is missing. Thanks!

authz/providers/azure/rbac/rbac.go
```diff
@@ -373,6 +377,11 @@ func (a *AccessInfo) CheckAccess(request *authzv1.SubjectAccessReviewSpec) (*aut
 }
 
 func (a *AccessInfo) sendCheckAccessRequest(ctx context.Context, checkAccessUsername string, checkAccessURL url.URL, checkAccessBody *CheckAccessRequest, ch chan *authzv1.SubjectAccessReviewStatus) error {
+	var clientInCtx *http.Client
+	if c, ok := ctx.Value(oauth2.HTTPClient).(*http.Client); ok {
```
bcho, line 381: nit: can we move this client loader to `util/azure` so we can hide the context key from caller?

authz/providers/azure/azure_test.go
```diff
@@ -37,7 +37,8 @@ import (
 )
 
 const (
-	loginResp = `{ "token_type": "Bearer", "expires_in": 8459, "access_token": "%v"}`
+	loginResp            = `{ "token_type": "Bearer", "expires_in": 8459, "access_token": "%v"}`
+	httpClientRetryCount = 2
```
bcho, line 41: can we add a unit test for covering the retry behavior?

authz/providers/azure/azure_test.go
```diff
@@ -52,9 +53,10 @@ func clientSetup(serverUrl, mode string) (*Authorizer, error) {
 	}
 
 	authOpts := auth.Options{
-		ClientID:     "client_id",
-		ClientSecret: "client_secret",
-		TenantID:     "tenant_id",
+		ClientID:             "client_id",
+		ClientSecret:         "client_secret",
+		TenantID:             "tenant_id",
+		HttpClientRetryCount: httpClientRetryCount,
```
julienstroheker, line 59: Can we assert that it try 2 times ?

authz/providers/azure/azure_test.go
```diff
@@ -171,6 +174,7 @@ func TestCheck(t *testing.T) {
 		assert.Nilf(t, resp, "response should be nil")
 		assert.NotNilf(t, err, "should get error")
 		assert.Contains(t, err.Error(), "Error occured during authorization check")
+		assert.Contains(t, err.Error(), fmt.Sprintf("giving up after %d attempt", httpClientRetryCount+1))
```
bingosummer, line 177: @bcho @julienstroheker this line can assert that it gives up after 3 attempts.

authz/providers/azure/azure_test.go
bingosummer: added

authz/providers/azure/azure_test.go
bingosummer: could you please have another round of review? Thanks

bcho (approved):
LGTM

bingosummer:
@weinong could you please have another round of review? Thanks

weinong:
@julienstroheker PTAL

weinong (approved):
approved

PR #374, opened 2023-09-19
cache authinfo since it won't change

weinong:
We don't need to call AAD discovery endpoint for every request.

### DIFF
--- a/auth/providers/azure/azure.go
+++ b/auth/providers/azure/azure.go
@@ -24,6 +24,7 @@ import (
 	"log"
 	"net/http"
 	"strings"
+	"sync"
 	"time"
 
 	"go.kubeguard.dev/guard/auth"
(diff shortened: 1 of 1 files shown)


auth/providers/azure/azure.go
```diff
@@ -79,19 +80,35 @@ type authInfo struct {
...
 		Options: opts,
 	}
-	authInfoVal, err := getAuthInfo(ctx, c.Environment, c.TenantID, c.HttpClientRetryCount, getMetadata)
-	if err != nil {
-		return nil, err
+
+	if cachedAuthInfo == nil {
+		mutex.Lock()
+		defer mutex.Unlock()
+		if cachedAuthInfo == nil {
```
vpatelsj, line 97: whats the reason for checking cachedAuthInfo == nil twice?

weinong: making sure it's really nil once it gets the mutex because potentially it's already set by other thread once this mutex is acquired.

auth/providers/azure/azure.go
```diff
@@ -79,19 +80,35 @@ type authInfo struct {
...
+			// getAuthInfo gets the issuer, graph, and AAD endpoints from the Azure metadata endpoint
+			// since it won't change at all, we can cache it
+			authInfoVal, err := getAuthInfo(ctx, c.Environment, c.TenantID, c.HttpClientRetryCount, getMetadata)
```
vpatelsj, line 100: If getAuthInfo returns an error constantly, cache will stay nil forever. Does that then lead to throttling?

weinong: yeah, but if it cannot really get result from it, there is no point to proceed.

PR #375, opened 2023-09-19
cache authinfo since it won't change

gambtho:
We don't need to call AAD discovery endpoint for every request.


weinong (approved):
LGTM, of course!

PR #376, opened 2023-09-20
fix: cache oidc issuer provider

bcho:


auth/providers/azure/azure.go
```diff
@@ -80,11 +80,48 @@ type authInfo struct {
...
+// TODO: combine auth info and issuer into one
 var (
 	cachedAuthInfo *authInfo
 	mutex          = &sync.Mutex{}
+
+	cachedOIDCIssuerProviders map[string]*oidc.Provider = make(map[string]*oidc.Provider)
+	cachedOIDCIssuerProvidersMutex = &sync.RWMutex{}
 )
 
+func getOIDCIssuerProvider(issuerURL string, issuerGetRetryCount int) (*oidc.Provider, error) {
```
weinong, line 92: this issuer really wont change

bcho: updated, ptal

auth/providers/azure/azure.go
```diff
@@ -80,11 +80,55 @@ type authInfo struct {
...
+	if cached, ok := getCachedOIDCIssuerProvider(); ok {
+		return cached, nil
+	}
+
+	// slow path: construct from remote
+	// NOTE: we hold the lock even it's doing HTTP call to avoid sending multiple requests
+	cachedOIDCIssuerProvidersMutex.Lock()
+	defer cachedOIDCIssuerProvidersMutex.Unlock()
+
+	if cached, ok := getCachedOIDCIssuerProvider(); ok {
```
jwilder, line 113: This will deadlock.  It tries to take an RLock while a Lock is already held.

jwilder (changes requested):

auth/providers/azure/azure.go
bcho: yes... sorry about the rush, writing a test in local

auth/providers/azure/azure.go
```diff
@@ -80,11 +80,58 @@ type authInfo struct {
...
+
+	if cached, ok := getCachedOIDCIssuerProviderUnsafe(); ok {
+		// another goroutine has already constructed the provider
+		return cached, nil
+	}
+
+	// NOTE: we start a root context here to allow background remote key set refresh
+	ctx := context.Background()
+	ctx = withRetryableHttpClient(ctx, issuerGetRetryCount)
+	fmt.Println(issuerURL)
```
weinong: debug? use klog instead

auth/providers/azure/azure.go
```diff
@@ -80,11 +80,58 @@ type authInfo struct {
...
```
jwilder: Remove?

bcho: removed!

jwilder (approved):

bcho:
I also added a simple test case to simulate the concurrent access, tested with `-race` from local:

```
go test -v -race ./auth/providers/azure
```

All tests passed.

weinong (approved):
LGTM, thanks!

weinong (approved):
LGTM

PR #378, opened 2023-09-20
log token claims at verbose level when failing token verification

gambtho:
log token claims at verbose level when failing token verification.
this may enable debugging on the source of the token


weinong:
@ tamalsaha can you approve this change?

PR #392, opened 2024-06-11
Fix idtyp issue with GetMemberGroupsUsingARCOboService

vineeth-thumma:
- During testing authentication flow in Guard for a user token (with overage claims), noticed that it is treated as SPN/Application and failed the authentication

- Looks like Microsoft identity platform has changed the format for user tokens and is now setting optional "idtyp" claim. This needs a change in Guard on how we evaluate if the token corresponds to an AAD User or an Application (SPN)

For more context,
"idtyp" claim is used to get the type of token (app, user, device). By default, it's only emitted for app-only tokens (which seems to have changed now)
Like all optional claims that affect the access token, the resource in the request must set this optional claim, since resources own the access token
https://learn.microsoft.com/en-us/entra/identity-platform/optional-claims-reference#additionalproperties-of-optional-claims

### DIFF
--- a/auth/providers/azure/graph/graph.go
+++ b/auth/providers/azure/graph/graph.go
@@ -202,8 +202,17 @@ func (u *UserInfo) getMemberGroupsUsingARCOboService(ctx context.Context, access
 		return nil, errors.Wrap(err, "Error while parsing accessToken for validation")
 	}
 
-	// the arc obo service does not support getting groups for applications
-	if claims[idtypClaim] != nil && !strings.EqualFold(claims[idtypClaim].(string), "user") {
+	// the arc obo service does not support getting groups for applications(SPN)
+	isAADUser := false
+	if (claims[idtypClaim] == nil) {
+		isAADUser = true
+	} else {
+		idtyp, ok := claims[idtypClaim].(string)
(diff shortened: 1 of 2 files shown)


auth/providers/azure/graph/graph.go
```diff
@@ -203,7 +203,7 @@ func (u *UserInfo) getMemberGroupsUsingARCOboService(ctx context.Context, access
 	}
 
 	// the arc obo service does not support getting groups for applications
-	if claims[idtypClaim] != nil {
+	if claims[idtypClaim] != nil && !strings.EqualFold(claims[idtypClaim].(string), "user") {
```
julienstroheker, line 206: What if `claims[idtypClaim]` is not a string somehow ? Can you add a UT to make sure we don't panic ?

vineeth-thumma: the logic has been updated, PTAL

amanohar (approved):

weinong (dismissed):
LGTM

bcho (approved):

julienstroheker (approved):

weinong (approved):
LGTM

PR #398, opened 2024-11-28
[BUG/FIX] Azure RBAC fix refresh token logic

saisankargochhayat:
This PR aims to fix the refresh token logic for Azure RBAC.
The existing refresh token logic relies on the `expires_in` value from the token response -
{"access_token":"<toeken>","expires_in":"86700","refresh_token":"","expires_on":"1732881796","not_before":"1732795096","resource":"https://management.azure.com","token_type":"Bearer"}
However in the case when the token is not fresh and returned from cache, the `expires_in` value misleads Guard to think the token doesn't need refreshing since it adds the current_time + expires_in to set the new expiry time.
In this PR we directly use the `expires_on` to set the expiry time accurately.

The PR also increases the tokenExpiryDelta to 5 minutes to ensure the token is refreshed 5 mins earlier than actual expiry. This is done on both AuthZ (rbac.go) and AuthN (graph.go) flow.

Removed `expires_in` to be returned from msi_adapter in favor of `expires_on`

Using the current implementation guard now sets the expires property correctly, i.e. knows it expires at 2024-11-29 21:05:56 +0000 UTC in the below example rather than doing - current-time + 86700 

Guard log - 
```
│ I1202 20:38:40.019063       1 graph.go:331] Token received with expires_in 86700 and expires_at 1733258179            

││ I1202 20:38:40.019114       1 graph.go:337] Token refreshed successfully at 2024-12-02 20:38:40.019110448 +0000 UTC m=+444.823514368. Expire at set to: 2024-12-03 20:31:19 +0000 UTC 
 ```
 
 Tested KAP flow - 
 
image

RBAC non proxy flow - 
image

### DIFF
--- a/auth/providers/azure/graph/graph.go
+++ b/auth/providers/azure/graph/graph.go
@@ -72,6 +72,7 @@ var (
 )
 
 const (
+	// Time delta to refresh token before expiry
 	tokenExpiryDelta       = 300 * time.Second
 	getMemberGroupsTimeout = 23 * time.Second
 	getterName             = "ms-graph"
--- a/auth/providers/azure/graph/graph.go
+++ b/auth/providers/azure/graph/graph.go
@@ -72,7 +72,7 @@ var (
 )
 
 const (
-	expiryDelta            = 60 * time.Second
+	expiryDelta            = 300 * time.Second
 	getMemberGroupsTimeout = 23 * time.Second
 	getterName             = "ms-graph"
 	arcAuthMode            = "arc"
@@ -327,11 +327,13 @@ func (u *UserInfo) RefreshToken(ctx context.Context, token string) error {
 		if err != nil {
 			return errors.Errorf("%s: failed to refresh token: %s", u.tokenProvider.Name(), err)
 		}
+		klog.Infof("Token received with expires_in %d and expires_at %d", resp.Expires, resp.ExpiresOn)
 		// Set the authorization headers for future requests
 		u.headers.Set("Authorization", fmt.Sprintf("Bearer %s", resp.Token))
-		expIn := time.Duration(resp.Expires) * time.Second
-		u.expires = time.Now().Add(expIn - expiryDelta)
-		klog.Infof("Token refreshed successfully on %s. Expire at:%s", time.Now(), u.expires)
+		// Use ExpiresOn to set the expiration time
+		expOn := time.Unix(int64(resp.ExpiresOn), 0)
+		u.expires = expOn.Add(-expiryDelta)
+		klog.Infof("Token refreshed successfully on %s. Expire at: %s", time.Now(), u.expires)
 	}
 
 	return nil
--- a/auth/providers/azure/graph/graph.go
+++ b/auth/providers/azure/graph/graph.go
@@ -72,7 +72,7 @@ var (
 )
 
 const (
-	expiryDelta            = 300 * time.Second
+	tokenExpiryDelta       = 300 * time.Second
 	getMemberGroupsTimeout = 23 * time.Second
 	getterName             = "ms-graph"
 	arcAuthMode            = "arc"
@@ -332,7 +332,7 @@ func (u *UserInfo) RefreshToken(ctx context.Context, token string) error {
 		u.headers.Set("Authorization", fmt.Sprintf("Bearer %s", resp.Token))
 		// Use ExpiresOn to set the expiration time
 		expOn := time.Unix(int64(resp.ExpiresOn), 0)
-		u.expires = expOn.Add(-expiryDelta)
+		u.expires = expOn.Add(-tokenExpiryDelta)
 		klog.Infof("Token refreshed successfully at %s. Expire at set to: %s", time.Now(), u.expires)
 	}
 
--- a/auth/providers/azure/graph/graph.go
+++ b/auth/providers/azure/graph/graph.go
@@ -327,7 +327,7 @@ func (u *UserInfo) RefreshToken(ctx context.Context, token string) error {
 		if err != nil {
 			return errors.Errorf("%s: failed to refresh token: %s", u.tokenProvider.Name(), err)
 		}
-		klog.Infof("Token received with expires_in %d and expires_at %d", resp.Expires, resp.ExpiresOn)
+		klog.Infof("Token received with expires_in %d and expires_at %d", resp.ExpiresIn, resp.ExpiresOn)
 		// Set the authorization headers for future requests
 		u.headers.Set("Authorization", fmt.Sprintf("Bearer %s", resp.Token))
 		// Use ExpiresOn to set the expiration time
--- a/auth/providers/azure/graph/graph.go
+++ b/auth/providers/azure/graph/graph.go
@@ -328,7 +328,7 @@ func (u *UserInfo) RefreshToken(ctx context.Context, token string) error {
 		if err != nil {
 			return errors.Errorf("%s: failed to refresh token: %s", u.tokenProvider.Name(), err)
 		}
-		klog.Infof("Token received with expires_in %d and expires_at %d", resp.ExpiresIn, resp.ExpiresOn)
+		klog.Infof("Token received, expires_at %d", resp.ExpiresOn)
 		// Set the authorization headers for future requests
 		u.headers.Set("Authorization", fmt.Sprintf("Bearer %s", resp.Token))
 		// Use ExpiresOn to set the expiration time
--- a/auth/providers/azure/graph/graph.go
+++ b/auth/providers/azure/graph/graph.go
@@ -333,7 +333,7 @@ func (u *UserInfo) RefreshToken(ctx context.Context, token string) error {
 		// Use ExpiresOn to set the expiration time
 		expOn := time.Unix(int64(resp.ExpiresOn), 0)
 		u.expires = expOn.Add(-expiryDelta)
-		klog.Infof("Token refreshed successfully on %s. Expire at: %s", time.Now(), u.expires)
+		klog.Infof("Token refreshed successfully at %s. Expire at set to: %s", time.Now(), u.expires)
 	}
 
 	return nil
--- a/auth/providers/azure/graph/graph_test.go
+++ b/auth/providers/azure/graph/graph_test.go
@@ -110,8 +110,8 @@ func TestLogin(t *testing.T) {
   "access_token": "%s",
   "expires_on": %d
 }`
-		expiresOn := time.Now().Add(time.Second * 3599).Unix()
-		ts, u := getAuthServerAndUserInfo(http.StatusOK, fmt.Sprintf(validBody, validToken, expiresOn), "jason", "bourne")
+		expiresOn := time.Now().Add(time.Second * 3599)
+		ts, u := getAuthServerAndUserInfo(http.StatusOK, fmt.Sprintf(validBody, validToken, expiresOn.Unix()), "jason", "bourne")
 		defer ts.Close()
 		err := u.RefreshToken(ctx, "")
 		if err != nil {
@@ -122,6 +122,10 @@ func TestLogin(t *testing.T) {
 		}
 		if !time.Now().Before(u.expires) {
 			t.Errorf("Expiry not set properly. Expected it to be after the current time. Actual: %v", u.expires)
+		}
+		// Expect u.expires to be tokenExpiryDelta before expiresOn
+		if expiresOn.Add(-tokenExpiryDelta).Equal(u.expires) {
+			t.Errorf("Expiry not set properly. Expected it to be %d before expiresOn. Actual: %v", tokenExpiryDelta, u.expires)
 		}
 	})
 
--- a/auth/providers/azure/graph/graph_test.go
+++ b/auth/providers/azure/graph/graph_test.go
@@ -107,11 +107,12 @@ func TestLogin(t *testing.T) {
 		validBody := `{
   "token_type": "Bearer",
   "expires_in": 3599,
-  "access_token": "%s"
+  "access_token": "%s",
+  "expires_on": %d
 }`
-		ts, u := getAuthServerAndUserInfo(http.StatusOK, fmt.Sprintf(validBody, validToken), "jason", "bourne")
+		expiresOn := time.Now().Add(time.Second * 3599).Unix()
+		ts, u := getAuthServerAndUserInfo(http.StatusOK, fmt.Sprintf(validBody, validToken, expiresOn), "jason", "bourne")
 		defer ts.Close()
-
 		err := u.RefreshToken(ctx, "")
 		if err != nil {
 			t.Errorf("Error when trying to log in: %s", err)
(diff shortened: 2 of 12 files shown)

### CODE REVIEW

File: auth/providers/azure/graph/graph.go [hunk 1]
```diff
@@ -72,7 +72,7 @@ var (
 )
 
 const (
-	expiryDelta            = 60 * time.Second
+	expiryDelta            = 300 * time.Second
```
vineeth-thumma, line 75: rename to `tokenExpiryDelta` to be specific?

vineeth-thumma (changes requested):

auth/providers/azure/graph/graph.go
saisankargochhayat: This is not a new var, but I can add a comment?

auth/providers/azure/graph/graph.go
```diff
@@ -327,11 +327,13 @@ func (u *UserInfo) RefreshToken(ctx context.Context, token string) error {
...
+		klog.Infof("Token received with expires_in %d and expires_at %d", resp.Expires, resp.ExpiresOn)
 		// Set the authorization headers for future requests
 		u.headers.Set("Authorization", fmt.Sprintf("Bearer %s", resp.Token))
-		expIn := time.Duration(resp.Expires) * time.Second
-		u.expires = time.Now().Add(expIn - expiryDelta)
-		klog.Infof("Token refreshed successfully on %s. Expire at:%s", time.Now(), u.expires)
+		// Use ExpiresOn to set the expiration time
+		expOn := time.Unix(int64(resp.ExpiresOn), 0)
+		u.expires = expOn.Add(-expiryDelta)
+		klog.Infof("Token refreshed successfully on %s. Expire at: %s", time.Now(), u.expires)
```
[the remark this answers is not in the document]
saisankargochhayat: Updated to at

auth/providers/azure/graph/types.go
```diff
@@ -21,6 +21,8 @@ type AuthResponse struct {
 	TokenType string `json:"token_type"`
 	Expires   int    `json:"expires_in"`
```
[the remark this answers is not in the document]
saisankargochhayat: Renamed to ExpiresIn

auth/providers/azure/graph/graph.go
saisankargochhayat: Updated to suggested name

auth/providers/azure/graph/graph.go
```diff
@@ -72,7 +72,8 @@ var (
 )
 
 const (
-	expiryDelta            = 60 * time.Second
+	// Time delta to refresh token before expiry
+	tokenExpiryDelta       = 300 * time.Second
```
AzureMarker, line 76: Can you also call out this change from 1min to 5min in the PR description?

auth/providers/azure/graph/graph_test.go
```diff
@@ -107,11 +107,12 @@ func TestLogin(t *testing.T) {
 		validBody := `{
   "token_type": "Bearer",
   "expires_in": 3599,
-  "access_token": "%s"
+  "access_token": "%s",
+  "expires_on": %d
 }`
-		ts, u := getAuthServerAndUserInfo(http.StatusOK, fmt.Sprintf(validBody, validToken), "jason", "bourne")
+		expiresOn := time.Now().Add(time.Second * 3599).Unix()
+		ts, u := getAuthServerAndUserInfo(http.StatusOK, fmt.Sprintf(validBody, validToken, expiresOn), "jason", "bourne")
```
AzureMarker, line 114: Can you add a check to make sure the expiration time is exactly the `expiresOn` value minus 5min?
Currently the test doesn't validate the change in behavior.

auth/providers/azure/graph/msi_tokenprovider_test.go
```diff
@@ -52,13 +53,25 @@ func TestMSITokenProvider(t *testing.T) {
...
 
 		if resp.Token != msiAccessToken {
 			t.Errorf("returned obo token '%s' doesn't match expected '%s'", resp.Token, msiAccessToken)
 		}
 		if resp.TokenType != expectedTokenType {
 			t.Errorf("expected token type: Bearer, actual: %s", resp.TokenType)
 		}
+
+		expectedExpiresOn, _ := strconv.Atoi(expectedResp.ExpiresOn)
+		if resp.ExpiresOn != expectedExpiresOn {
```
AzureMarker, line 72: To simplify this, can you have the expected `expires_on` value be a constant that gets used in the mock token response and here?

authz/providers/azure/rbac/rbac.go
```diff
@@ -49,13 +49,14 @@ import (
...
-	remainingSubReadARMHeader  = "x-ms-ratelimit-remaining-subscription-reads"
-	expiryDelta                = 60 * time.Second
+	managedClusters           = "Microsoft.ContainerService/managedClusters"
+	fleets                    = "Microsoft.ContainerService/fleets"
+	connectedClusters         = "Microsoft.Kubernetes/connectedClusters"
+	checkAccessPath           = "/providers/Microsoft.Authorization/checkaccess"
+	checkAccessAPIVersion     = "2018-09-01-preview"
+	remainingSubReadARMHeader = "x-ms-ratelimit-remaining-subscription-reads"
+	// Time delta to refresh token before expiry
+	tokenExpiryDelta           = 300 * time.Second
```
AzureMarker, line 59: Also call out that the time delta changed for authz.

authz/providers/azure/rbac/rbac_test.go
```diff
@@ -178,9 +178,11 @@ func TestLogin(t *testing.T) {
 		validBody := `{
 							"token_type": "Bearer",
 							"expires_in": 3599,
-							"access_token": "%s"
+							"access_token": "%s",
+							"expires_on": %d
 						}`
-		ts, u := getAuthServerAndAccessInfo(http.StatusOK, fmt.Sprintf(validBody, validToken), "jason", "bourne")
+		expiresOn := time.Now().Add(time.Second * 3599).Unix()
+		ts, u := getAuthServerAndAccessInfo(http.StatusOK, fmt.Sprintf(validBody, validToken, expiresOn), "jason", "bourne")
```
AzureMarker, line 185: This test should validate the new behavior.

auth/providers/azure/graph/graph.go
saisankargochhayat: Updated

authz/providers/azure/rbac/rbac.go
saisankargochhayat: Updated

auth/providers/azure/graph/msi_tokenprovider_test.go
saisankargochhayat: Updated based on suggestion

auth/providers/azure/graph/graph_test.go
saisankargochhayat: Updated to test the expected behavior

authz/providers/azure/rbac/rbac_test.go
saisankargochhayat: Updated to test the expected behavior

auth/providers/azure/graph/msi_tokenprovider_test.go
```diff
@@ -28,9 +28,10 @@ func TestMSITokenProvider(t *testing.T) {
 	const (
 		inputAccessToken    = "inputAccessToken"
 		msiAccessToken      = "msiAccessToken"
-		tokenResponse       = `{"token_type":"Bearer","expires_in":"3599","access_token":"%s"}`
+		tokenResponse       = `{"access_token":"%s","expires_in":"86700","refresh_token":"","expires_on":"1732881796","not_before":"%d","resource":"https://management.azure.com","token_type":"Bearer"}`
```
AzureMarker, line 31: I think you put the `%d` on the wrong property?

saisankargochhayat: Nice catch thanks!

auth/providers/azure/graph/msi_tokenprovider.go
```diff
@@ -96,9 +96,15 @@ func (u *msiTokenProvider) Acquire(ctx context.Context, token string) (AuthRespo
 	}
 
 	authResp.TokenType = tokenResp.TokenType
-	authResp.Expires, err = strconv.Atoi(tokenResp.ExpiresIn)
+	// This value is true only at token creation time, if a cached token is used, its not accurate
+	authResp.ExpiresIn, err = strconv.Atoi(tokenResp.ExpiresIn)
```
julienstroheker, line 100: IMO if we don't use this, we should remove it from the parsed responsed json payload to avoid using it elsewhere

auth/providers/azure/graph/types.go
```diff
@@ -19,8 +19,10 @@ package graph
 // AuthResponse represents a response from the MS Graph auth API
 type AuthResponse struct {
 	TokenType string `json:"token_type"`
-	Expires   int    `json:"expires_in"`
+	ExpiresIn int    `json:"expires_in"`
```
julienstroheker, line 22: Like mentioned we should remove this

julienstroheker (commented):
LGTM, I'll let you decide if you want to apply my comment about the returned payload from MSI adapter

julienstroheker:
Also rename the PR with [BUG/FIX] in the title

auth/providers/azure/graph/graph_test.go
```diff
@@ -107,11 +107,12 @@ func TestLogin(t *testing.T) {
 		validBody := `{
   "token_type": "Bearer",
   "expires_in": 3599,
-  "access_token": "%s"
+  "access_token": "%s",
```
julienstroheker, line 110: you should then update this paylod to remove the `expired_in` and anywhere else

saisankargochhayat: Updated

auth/providers/azure/graph/types.go
saisankargochhayat: Updated

vineeth-thumma:
LGTM!

auth/providers/azure/graph/graph_test.go
```diff
@@ -122,6 +122,10 @@ func TestLogin(t *testing.T) {
 		if !time.Now().Before(u.expires) {
 			t.Errorf("Expiry not set properly. Expected it to be after the current time. Actual: %v", u.expires)
 		}
+		// Expect u.expires to be tokenExpiryDelta before expiresOn
+		if expiresOn.Add(-tokenExpiryDelta).Equal(u.expires) {
```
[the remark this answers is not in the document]
saisankargochhayat: Thanks fixed

authz/providers/azure/rbac/rbac_test.go
```diff
@@ -194,6 +195,10 @@ func TestLogin(t *testing.T) {
 		if !time.Now().Before(u.expiresAt) {
 			t.Errorf("Expiry not set properly. Expected it to be after the current time. Actual: %v", u.expiresAt)
 		}
+		// Expect u.expires to be tokenExpiryDelta before expiresOn
+		if expiresOn.Add(-tokenExpiryDelta).Equal(u.expiresAt) {
```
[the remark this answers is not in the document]
saisankargochhayat: Thanks fixed

julienstroheker (approved):
LGTM on Arc Side but need review on AKS Side since we are touching the token interface.

auth/providers/azure/graph/graph_test.go
```diff
@@ -122,9 +122,12 @@ func TestLogin(t *testing.T) {
 		if !time.Now().Before(u.expires) {
 			t.Errorf("Expiry not set properly. Expected it to be after the current time. Actual: %v", u.expires)
 		}
+		// Normalize to second precision for comparison
+		expectedExpiresOn := expiresOn.Add(-tokenExpiryDelta).Truncate(time.Second)
+		actualExpires := u.expires.Truncate(time.Second)
 		// Expect u.expires to be tokenExpiryDelta before expiresOn
```
vineeth-thumma, line 128: this comment is not relevant anymore? or might be confusing given that you're normalizing it above for comparision

authz/providers/azure/rbac/rbac_test.go
```diff
@@ -195,9 +195,13 @@ func TestLogin(t *testing.T) {
 		if !time.Now().Before(u.expiresAt) {
 			t.Errorf("Expiry not set properly. Expected it to be after the current time. Actual: %v", u.expiresAt)
 		}
+
+		// Normalize to second precision for comparison
+		expectedExpiresOn := expiresOn.Add(-tokenExpiryDelta).Truncate(time.Second)
+		actualExpires := u.expiresAt.Truncate(time.Second)
 		// Expect u.expires to be tokenExpiryDelta before expiresOn
```
vineeth-thumma, line 202: this comment is not relevant anymore this comment is not relevant anymore? or might be confusing given that you're normalizing it above for comparison

vineeth-thumma (changes requested):

vineeth-thumma (approved):
LGTM

authz/providers/azure/rbac/rbac.go
```diff
@@ -224,9 +225,11 @@ func (a *AccessInfo) RefreshToken(ctx context.Context) error {
 
 		// Set the authorization headers for future requests
 		a.headers.Set("Authorization", fmt.Sprintf("Bearer %s", resp.Token))
-		expIn := time.Duration(resp.Expires) * time.Second
-		a.expiresAt = time.Now().Add(expIn - expiryDelta)
-		klog.Infof("Token refreshed successfully on %s. Expire at:%s", time.Now(), a.expiresAt)
+
+		// Use ExpiresOn to set the expiration time
+		expOn := time.Unix(int64(resp.ExpiresOn), 0)
+		a.expiresAt = expOn.Add(-tokenExpiryDelta)
```
bcho, line 231: Will the tokenProvider.Acquire guarantee to refresh the token or it's possible to load from a cache? In the latter case, will there be chance that the `resp.ExpiresOn` is less than 5 mins from time.Now(), which causes the a.expiresAt < time.Now() ? What's the impact of seeing this, like will that cause the RefreshToken being call repeatedly and eventualy cause busy loop?

saisankargochhayat: So, from the tokenProvider.Acquire point of view, it's always going to get a token with at-least 1 hour of expiry left
As the identity controller always refreshes the token 1 hour before expiry - 
https://github.com/azure-core/ClusterConfigurationAgent/blob/9e24b19ffdc7f71f53cb4fb86cb3d7d830be6856/Utils/TokenProviders/ManagedIdentityTokenProvider.go#L78

So, we should never hit a scenario where the ExpiresOn is less than 5mins from time.Now()

bcho (approved):

authz/providers/azure/rbac/rbac.go
julienstroheker: For context, in Arc we have on our own `msi-adapter` component that mimic IMDS on the edge. Like @saisankargochhayat it just fetch a "cached" token from another component "identity-controller" that manage getting token and save it as k8s secret via a CR. The code that @saisankargochhayat linked is where `msi-adapter` decide to trigger a token rotation of re-used the cache secret within the 1H expiration.

julienstroheker (approved):

weinong (approved):
LGTM

PR #409, opened 2025-03-15
Azure RBAC support for CRD authorization

BrunoMPires:
AKS resource provider API version 2025-02-01 introduces `Microsoft.ContainerService/managedClusters/customresources` set of DataActions to enable athorizing `/read`, `/write` and `/delete` actions on customer CRs via Azure RBAC custom roles. As stated in [before you begin](https://learn.microsoft.com/en-us/azure/aks/manage-azure-rbac?tabs=azure-cli#before-you-begin) this was not previously supported.

When providing access to CRs through `Microsoft.ContainerService/managedClusters/customresources` CheckAccess API request must contain `body/group`, `body/version` and `body/resource` action attributes to allow ABAC conditions evaluation.

This change enables Guard to classify resources that don't belong to any API group recognized by Azure RBAC as a CR. In this case Guard will make the CheckAccessAPI request using `Microsoft.ContainerService/managedClusters/customresources` DataAction and include the resource's group, version and kind as action attributes.


weinong:
can you add a flag to enable this new behavior so that we can control how we would like to phase in the CRD support?

authz/providers/azure/rbac/checkaccessreqhelper.go
```diff
@@ -127,6 +127,29 @@ type AuthorizationDecision struct {
 	TimeToLiveInMs      int                 `json:"timeToLiveInMs"`
 }
 
+func getCustomResourceOperationsMap(clusterType string) map[string]azureutils.AuthorizationActionInfo {
+	return map[string]azureutils.AuthorizationActionInfo{
+		"read": {
+			IsDataAction: true,
+			AuthorizationEntity: azureutils.AuthorizationEntity{
+				Id: path.Join(clusterType, "customresources", "read"),
```
weinong, line 135: use consts for common strings

BrunoMPires:
> can you add a flag to enable this new behavior so that we can control how we would like to phase in the CRD support?

@weinong I added `AllowCustomResourceTypeCheck` flag to control the feature. When disabled the behavior will be same as before these changes were introduced.

authz/providers/azure/rbac/checkaccessreqhelper.go
```diff
@@ -230,6 +258,20 @@ func getDataActions(subRevReq *authzv1.SubjectAccessReviewSpec, clusterType stri
 				IsDataAction: true,
 			}
 
+			if !allowCustomResourceTypeCheck || storedOperationsMap == nil || len(storedOperationsMap) == 0 {
```
arxhive, line 261: It's a bit complicated. Let's improve readability of this if / else statement. For instance, extract a well named variable something like:
```
isCustomerResourceTypeCheckAvailable := !allowCustomResourceTypeCheck || storedOperationsMap == nil || len(storedOperationsMap) == 0
```

The second else if part is even more complicated. maybe "resourceIsFound"? Or what exactly you're trying to emphasize there?
```

File: authz/providers/azure/rbac/checkaccessreqhelper.go [hunk 3]
```diff
@@ -230,6 +258,20 @@ func getDataActions(subRevReq *authzv1.SubjectAccessReviewSpec, clusterType stri
...
+				/*
+					In this case, CustomResource type verification is not available. Either the storedOperationsMap is empty or feature is disabled.
+				*/
+				klog.V(7).Info("CustomResource type verification is not available for this request")
+			} else if _, found := storedOperationsMap[subRevReq.ResourceAttributes.Group][subRevReq.ResourceAttributes.Resource]; !found &&
+				subRevReq.ResourceAttributes.Group != "" && subRevReq.ResourceAttributes.Resource != "" {
+				/*
+					In this case both Res and Group are not *, but there is no matching DataAction present on the storedOperationsMap.
```
arxhive, line 269: So, why don't we respect `allowCustomResourceTypeCheck` flag for this part?

authz/providers/azure/rbac/checkaccessreqhelper.go
```diff
@@ -230,6 +258,20 @@ func getDataActions(subRevReq *authzv1.SubjectAccessReviewSpec, clusterType stri
 				IsDataAction: true,
 			}
 
+			if !allowCustomResourceTypeCheck || storedOperationsMap == nil || len(storedOperationsMap) == 0 {
+				/*
+					In this case, CustomResource type verification is not available. Either the storedOperationsMap is empty or feature is disabled.
+				*/
+				klog.V(7).Info("CustomResource type verification is not available for this request")
+			} else if _, found := storedOperationsMap[subRevReq.ResourceAttributes.Group][subRevReq.ResourceAttributes.Resource]; !found &&
+				subRevReq.ResourceAttributes.Group != "" && subRevReq.ResourceAttributes.Resource != "" {
```
arxhive, line 267: Since `Version` is a separate property of `ResourceAttributes` I think we should validate it as well.

authz/providers/azure/rbac/checkaccessreqhelper.go
```diff
@@ -286,6 +327,12 @@ func getAuthInfoListForWildcard(subRevReq *authzv1.SubjectAccessReviewSpec, stor
 		if subRevReq.ResourceAttributes.Group == "*" {
 			// all resources under all apigroups
 			filteredOperations = storedOperationsMap
+		} else if _, found := storedOperationsMap[subRevReq.ResourceAttributes.Group]; !found && subRevReq.ResourceAttributes.Group != "" && allowCustomResourceTypeCheck {
```
arxhive, line 330: Yet another piece deserved a bool variable imho.

arxhive, line 330: Yet another piece deserved a bool variable imho.

authz/providers/azure/rbac/checkaccessreqhelper.go
```diff
@@ -286,6 +327,12 @@ func getAuthInfoListForWildcard(subRevReq *authzv1.SubjectAccessReviewSpec, stor
...
+			/*
+				In this case Group is not *, but there are no matching DataActions present on the storedOperationsMap.
+				The resource is presumed to be a CR and <clusterType>/customresources/<action> DataAction will be used for check access.
```
arxhive, line 333: Don't we need to validate `ResourceAttributes.Version` again?

authz/providers/azure/rbac/checkaccessreqhelper.go
```diff
@@ -342,6 +389,13 @@ func getAuthInfoListForWildcard(subRevReq *authzv1.SubjectAccessReviewSpec, stor
 					finalFilteredOperations[group][subRevReq.ResourceAttributes.Resource] = verbMap
 				}
 			}
+		} else if _, found := storedOperationsMap[subRevReq.ResourceAttributes.Group][subRevReq.ResourceAttributes.Resource]; !found &&
+			subRevReq.ResourceAttributes.Group != "" && subRevReq.ResourceAttributes.Resource != "" && allowCustomResourceTypeCheck {
```
arxhive, line 393: Version

authz/providers/azure/rbac/checkaccessreqhelper.go
```diff
@@ -371,6 +425,49 @@ func getAuthInfoListForWildcard(subRevReq *authzv1.SubjectAccessReviewSpec, stor
...
+	return authInfoList, nil
+}
+
+func setAuthInfoResourceAttributes(action *azureutils.AuthorizationActionInfo, subRevReq *authzv1.SubjectAccessReviewSpec) error {
+	if subRevReq.ResourceAttributes == nil {
+		return errors.New("Resource attributes are empty")
+	}
+	if subRevReq.ResourceAttributes.Group == "" {
+		return errors.New("Group is empty")
+	}
```
arxhive, line 459: I'd return an error if there is no version as well

authz/providers/azure/rbac/checkaccessreqhelper_test.go
```diff
@@ -55,6 +61,31 @@ func createOperationsMap(clusterType string) azureutils.OperationsMap {
 	}
 }
 
+func addOperationMap(operationsMap azureutils.OperationsMap, subRevReq *authzv1.SubjectAccessReviewSpec, authInfo []azureutils.AuthorizationActionInfo) {
+	if subRevReq.ResourceAttributes != nil {
+		group := subRevReq.ResourceAttributes.Group
+		resource := subRevReq.ResourceAttributes.Resource
+		verb := subRevReq.ResourceAttributes.Verb
+
```
arxhive, line 69: Version?

authz/providers/azure/rbac/checkaccessreqhelper_test.go
```diff
@@ -55,6 +61,31 @@ func createOperationsMap(clusterType string) azureutils.OperationsMap {
 	}
 }
 
+func addOperationMap(operationsMap azureutils.OperationsMap, subRevReq *authzv1.SubjectAccessReviewSpec, authInfo []azureutils.AuthorizationActionInfo) {
```
arxhive, line 64: Can we add a brief commeng to describe this function? Do we map general data actions? CR Data Actions? Or both? Being frank, spent few minutes but couldn't understand this part of code.

authz/providers/azure/rbac/checkaccessreqhelper_test.go
BrunoMPires: @arxhive version is not included here because it is not a dimension of the `operationsMap`, which holds the existing DataActions fetched from Azure using the resource provider _/operations_ endpoint, in a three-dimensional structure: `operationsMap[group][resource][action]`. E.g: _Microsoft.ContainerService/managedClusters/apps/deployments/read_ is `operationsmap["apps"]["deployments"]["read"]`. In fact, version info is stripped out of the resource's GroupVersion string at [utils.go](https://github.com/kubeguard/guard/blob/efc33724370945d44de0afbf9d3d18757d2772fa/util/azure/utils.go#L319), when building the operationsMap. Most available Azure RBAC DataActions do not include API group version, adding such support would require a larger effort refactoring Guard, possibily restructuring the available DataActions.

authz/providers/azure/rbac/checkaccessreqhelper.go
BrunoMPires: @arxhive See my previous coment about version.

authz/providers/azure/rbac/checkaccessreqhelper.go
BrunoMPires: @arxhive previous comments about version also apply here.

authz/providers/azure/rbac/checkaccessreqhelper.go
BrunoMPires: @arxhive previous comments about version also apply here.

authz/providers/azure/rbac/checkaccessreqhelper.go
BrunoMPires: We verified in the _if_ portion of this statement. I refactored to improve clarity.

authz/providers/azure/rbac/checkaccessreqhelper.go
BrunoMPires: Right, thanks for the suggestion, updated.

authz/providers/azure/rbac/checkaccessreqhelper.go
BrunoMPires: Done.

authz/providers/azure/rbac/checkaccessreqhelper_test.go
BrunoMPires: I refactored the test so the funcation could be removed. Now I'm calling getDataActions using `tt.args.isCrTest`

authz/providers/azure/rbac/checkaccessreqhelper.go
```diff
@@ -373,6 +426,65 @@ func getAuthInfoListForWildcard(subRevReq *authzv1.SubjectAccessReviewSpec, stor
...
+	if subRevReq.ResourceAttributes == nil {
+		return false
+	}
+	if subRevReq.ResourceAttributes.Group == "" || subRevReq.ResourceAttributes.Group == "*" {
+		return false
+	}
+	if subRevReq.ResourceAttributes.Resource == "" || subRevReq.ResourceAttributes.Resource == "*" {
+		_, found := (*operationsMap)[subRevReq.ResourceAttributes.Group]
+		return !found
+	} else {
```
weinong, line 482: this else is unnecessary. imo, removing it will improve readability as what this function is doing is just
if
else if
else if
else

authz/providers/azure/rbac/checkaccessreqhelper.go
```diff
@@ -344,6 +391,12 @@ func getAuthInfoListForWildcard(subRevReq *authzv1.SubjectAccessReviewSpec, stor
 					finalFilteredOperations[group][subRevReq.ResourceAttributes.Resource] = verbMap
 				}
 			}
+		} else if isCustomerResourceTypeCheckAvailable && isCustomResourceAccessRequest(subRevReq, &storedOperationsMap) {
```
weinong, line 394: Custom Resources will be absent for scenarios in L388 and L400. Aren't they also valid scenarios?

authz/providers/azure/rbac/checkaccessreqhelper.go
weinong: Same question in https://github.com/kubeguard/guard/pull/409/files#r2132760865
L329 will miss the Custom Resource

authz/providers/azure/rbac/checkaccessreqhelper.go
```diff
@@ -214,24 +241,39 @@ func getResourceAndAction(resource string, subResource string, verb string) stri
...
+func getDataActions(subRevReq *authzv1.SubjectAccessReviewSpec, clusterType string, allowCustomResourceTypeCheck bool) ([]azureutils.AuthorizationActionInfo, error) {
 	var authInfoList []azureutils.AuthorizationActionInfo
 	var err error
 
 	if subRevReq.ResourceAttributes != nil {
+		storedOperationsMap := getStoredOperationsMap()
+
+		isCustomerResourceTypeCheckAvailable := allowCustomResourceTypeCheck && storedOperationsMap != nil && len(storedOperationsMap) != 0
+		if !isCustomerResourceTypeCheckAvailable {
+			klog.V(5).Info("CustomResource type verification is not available for this request")
```
julienstroheker, line 253: Is that not better to log only when available ?

authz/providers/azure/rbac/checkaccessreqhelper.go
```diff
@@ -373,6 +426,65 @@ func getAuthInfoListForWildcard(subRevReq *authzv1.SubjectAccessReviewSpec, stor
...
+
+func setAuthInfoResourceAttributes(action *azureutils.AuthorizationActionInfo, subRevReq *authzv1.SubjectAccessReviewSpec) error {
+	if subRevReq.ResourceAttributes == nil {
+		return errors.New("Resource attributes are empty")
+	}
+	if subRevReq.ResourceAttributes.Group == "" {
+		return errors.New("Group is empty")
+	}
+	action.Attributes = make(map[string]string)
+	action.Attributes["Microsoft.ContainerService/managedClusters/customResources:group"] = subRevReq.ResourceAttributes.Group
```
julienstroheker, line 462: What about other clusterType ? (Arc)

authz/providers/azure/rbac/checkaccessreqhelper.go
```diff
@@ -373,6 +426,65 @@ func getAuthInfoListForWildcard(subRevReq *authzv1.SubjectAccessReviewSpec, stor
...
+func getAuthInfoListForCustomResource(subRevReq *authzv1.SubjectAccessReviewSpec, clusterType string) ([]azureutils.AuthorizationActionInfo, error) {
+	var authInfoList []azureutils.AuthorizationActionInfo
+	if subRevReq.ResourceAttributes.Verb == "*" {
+		for _, action := range getCustomResourceOperationsMap(clusterType) {
+			authInfoList = append(authInfoList, action)
+		}
+	} else {
+		action := getActionName(subRevReq.ResourceAttributes.Verb)
+		authInfoSingle, found := getCustomResourceOperationsMap(clusterType)[action]
+		if !found {
```
weinong, line 438: This should be documented as one of the limitations that we need to vet with PM and eventually documented publicly.

authz/providers/azure/rbac/checkaccessreqhelper.go
```diff
@@ -373,6 +426,65 @@ func getAuthInfoListForWildcard(subRevReq *authzv1.SubjectAccessReviewSpec, stor
...
+	if subRevReq.ResourceAttributes == nil {
+		return errors.New("Resource attributes are empty")
+	}
+	if subRevReq.ResourceAttributes.Group == "" {
+		return errors.New("Group is empty")
+	}
+	action.Attributes = make(map[string]string)
+	action.Attributes["Microsoft.ContainerService/managedClusters/customResources:group"] = subRevReq.ResourceAttributes.Group
+	if subRevReq.ResourceAttributes.Version != "" {
+		action.Attributes["Microsoft.ContainerService/managedClusters/customResources:version"] = subRevReq.ResourceAttributes.Version
```
weinong, line 464: this will be my definitive blocking comment. we should not account for version. this is not accounted in k8s nor in current azure rbac.

BrunoMPires: @weinong version will only be verified if the user creates a condition explicitly checking it, otherwise passing it on to CheckAccess API has no effect. In Azure RBAC the `:version` attribute is published for AKS /CustomResources DataActions. If version is avaialble and Guard doesn't passes it forward ABAC will not function properly, conditions using `:version` will always fail.

weinong: my point is `:version` will be removed from Azure RBAC. so it's better to do it now rather than later.

util/azure/utils.go
```diff
@@ -158,6 +158,7 @@ type AuthorizationEntity struct {
 type AuthorizationActionInfo struct {
 	AuthorizationEntity
 	IsDataAction bool `json:"IsDataAction"`
+	Attributes   map[string]string
```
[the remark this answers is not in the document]
BrunoMPires: Adressed on latest commit.

authz/providers/azure/rbac/checkaccessreqhelper.go
arxhive: As we aligned with team, we filter out Arc cluster types by design to avoid any unexpected side effects (this feature PRD and design were only considered for managedClusters use cases).

We also work iteratively here, let's support managedClusters first, a then commit for Arc support in the next iteration.

authz/providers/azure/rbac/checkaccessreqhelper.go
```diff
@@ -373,6 +426,65 @@ func getAuthInfoListForWildcard(subRevReq *authzv1.SubjectAccessReviewSpec, stor
...
+	}
+	if subRevReq.ResourceAttributes.Resource != "" {
+		action.Attributes["Microsoft.ContainerService/managedClusters/customResources:kind"] = subRevReq.ResourceAttributes.Resource
+	}
+	return nil
+}
+
+func isCustomResourceAccessRequest(subRevReq *authzv1.SubjectAccessReviewSpec, operationsMap *azureutils.OperationsMap) bool {
```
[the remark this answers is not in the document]
BrunoMPires: Addressed on latest commit.

authz/providers/azure/rbac/checkaccessreqhelper.go
julienstroheker: Azure RBAC for guard is for Managed Clusters and ARCs.
What and when are the next iterations you are talking about ? 
All cluster types should be included day one to make sure the design and code fits without any refactor in the future. Cc @amanohar

julienstroheker: This comment is not resolved.

julienstroheker:
I’m expecting someone from Azure RBAC’s team to review as required reviewer.

authz/providers/azure/rbac/checkaccessreqhelper.go
arxhive: Does ARC-RP already support CRD data actions? Can we just map the same verbs to arc here?

arxhive: I mean we did not design to build a generic feature at this time, the goal to cover existing hot user use cases first, and generalize this approach later. We're also planning to open it only in Preview mode for now.

We've sync with the RBAC team on a better, more explicit, and more agile notation for native CRD support by RBAC  for GA. PRD is still in progress, implementation is planned right after this release. I'm afraid that change request to existing design will delay expected release dates for the internal customers.

I guess including Arc in v2 would be a great idea.

But if you think that support for Arc data actions require just updating two lines of code - of course we can do it.

weinong (dismissed):
LGTM. please fix the build failure

weinong (approved):
LGTM

PR #413, opened 2025-05-09
managed namespaces support

artunduman:
We created a new resource type Microsoft.ContainerService/managedClusters/managedNamespaces that is the ARM object that represents a namespace in cluster. This PR is to support azure rbac on that object, with the same dataActions as the /namespaces/ scope.

As discussed with (and approved by) the Azure RBAC team, this is implementing a logical OR between /namespaces/ and /managedNamespaces/ scopes.

The logical OR allows the user to create a managedNamespace that takes over an existing namespace and not break the legacy role assignments at the /namespaces/ scope.

We only enable this check for AKS clusters that have managed namespaces. We gate that with the new flag azure.use-managed-namespace-resource-scope-format. 

**This change shouldn't affect the existing behaviour for the clusters without managed namespaces.**

Some cases that I tested in standalone **after the refactor:**

Assignments are 
RBAC Reader on namespaces/ns1
RBAC Admin on managedNamespaces/ns2

image

Test cases before refactor:

No access to ns1
Admin on managedNamespaces/ns2
No namespaces/ access
￼
Pasted Graphic

Reader on namespaces/ns1
Admin on managedNamespaces/ns2
￼
Pasted Graphic 1

No access to any managedNamespaces or namespaces
Reader on cluster
￼
Pasted Graphic 2

Reader on managedNamespaces/ns1
￼
Pasted Graphic 3


authz/providers/azure/rbac/rbac.go
```diff
@@ -375,7 +378,94 @@ func (a *AccessInfo) CheckAccess(request *authzv1.SubjectAccessReviewSpec) (*aut
...
+	}
+
+	if !a.enableManagedNamespaceRBAC || a.clusterType != managedClusters {
+		klog.V(5).Infof("Checkaccess request is denied for user %s", checkAccessUsername)
+		return finalStatus, nil
+	}
+
+	klog.V(5).Infof("Falling back to checking managed namespace scope", checkAccessUsername)
+
+	checkAccessURLManagedNS := *a.apiURL
```
artunduman, line 393: this point on, pretty much a duplicate of the above to not break existing scopes

authz/providers/azure/options/options.go
```diff
@@ -47,6 +47,7 @@ type Options struct {
 	AllowNonResDiscoveryPathAccess      bool
 	UseNamespaceResourceScopeFormat     bool
 	DiscoverResources                   bool
+	EnableManagedNamespaceRBAC          bool
```
chengliangli0918, line 50: could we UseManagedNamespaceResourceScopeFormat to follow the existing variable pattern (UseNamespaceResourceScopeFormat)

chengliangli0918: also could we also initialize this new variable to false in NewOptions() at Ln 62

authz/providers/azure/rbac/rbac.go
```diff
@@ -375,7 +378,94 @@ func (a *AccessInfo) CheckAccess(request *authzv1.SubjectAccessReviewSpec) (*aut
...
+	eg, egCtx = errgroup.WithContext(ctx)
+
+	ch = make(chan *authzv1.SubjectAccessReviewStatus, len(checkAccessBodies))
+	if len(checkAccessBodies) > 1 {
+		klog.V(5).Infof("Number of checkaccess requests to make: %d", len(checkAccessBodies))
+	}
+	eg.SetLimit(len(checkAccessBodies))
+	for _, checkAccessBody := range checkAccessBodies {
+		body := checkAccessBody
+		body.Resource.Id = path.Join(a.azureResourceId, managedNameSpaceString)
```
chengliangli0918, line 419: Why we need this line?

artunduman: If the scope we are checking and the Resource.Id are mismatched, I get an error from ARM. That's why I had to update the Resource.Id

chengliangli0918 (approved):

authz/providers/azure/rbac/checkaccessreqhelper.go
```diff
@@ -538,6 +539,15 @@ func getNameSpaceScope(req *authzv1.SubjectAccessReviewSpec, useNamespaceResourc
 	return false, namespace
 }
 
+func getManagedNameSpaceScope(req *authzv1.SubjectAccessReviewSpec) (bool, string) {
```
julienstroheker, line 542: nit `getManagedNamespaceScope`

authz/providers/azure/rbac/rbac.go
```diff
@@ -375,7 +378,94 @@ func (a *AccessInfo) CheckAccess(request *authzv1.SubjectAccessReviewSpec) (*aut
...
+	if finalStatus != nil && finalStatus.Allowed {
+		klog.V(5).Infof("Checkaccess request is allowed for user %s", checkAccessUsername)
+		return finalStatus, nil
+	}
+
+	if !a.useManagedNamespaceResourceScopeFormat || a.clusterType != managedClusters {
+		klog.V(5).Infof("Checkaccess request is denied for user %s", checkAccessUsername)
+		return finalStatus, nil
+	}
+
```
julienstroheker, line 390: is there a scenario where we can land in this part of the code for a managed cluster with `useManagedNamespaceResourceScopeFormat` false ?

authz/providers/azure/rbac/rbac.go
```diff
@@ -375,7 +378,94 @@ func (a *AccessInfo) CheckAccess(request *authzv1.SubjectAccessReviewSpec) (*aut
...
+				err = errutils.WithCode(errors.Errorf("Error: %s. Correlation ID: %s", err, requestUUID.String()), code)
+				return err
+			}
+			return nil
+		})
+	}
+
+	if err := eg.Wait(); err != nil {
+		if ctx.Err() == context.DeadlineExceeded {
+			klog.V(5).Infof("Checkaccess requests have timed out. Error: %v", ctx.Err())
```
julienstroheker, line 440: Errorf

authz/providers/azure/rbac/rbac.go
```diff
@@ -375,7 +378,94 @@ func (a *AccessInfo) CheckAccess(request *authzv1.SubjectAccessReviewSpec) (*aut
...
+			return nil, err
+		}
+	}
+	close(ch)
+
+	var finalStatusManagedNS *authzv1.SubjectAccessReviewStatus
+
+	for status := range ch {
+		if status.Denied {
+			klog.V(5).Infof("Checkaccess request is denied because one of the actions is denied for user %s in managedNamespace scope", checkAccessUsername)
```
julienstroheker, line 461: Should this  be an Errorf?

authz/providers/azure/rbac/rbac.go
```diff
@@ -375,7 +378,94 @@ func (a *AccessInfo) CheckAccess(request *authzv1.SubjectAccessReviewSpec) (*aut
...
+		klog.V(5).Infof("Checkaccess request is allowed for user %s", checkAccessUsername)
+		return finalStatus, nil
+	}
+
+	if !a.useManagedNamespaceResourceScopeFormat || a.clusterType != managedClusters {
+		klog.V(5).Infof("Checkaccess request is denied for user %s", checkAccessUsername)
+		return finalStatus, nil
+	}
+
+	klog.V(5).Infof("Falling back to checking managed namespace scope", checkAccessUsername)
```
julienstroheker, line 391: Should we extract all this new logic to a new sub function ?

authz/providers/azure/rbac/checkaccessreqhelper.go
artunduman: I tried to keep it consistent with getNameSpaceScope method (used above)

authz/providers/azure/rbac/rbac.go
```diff
@@ -298,6 +300,7 @@ func (a *AccessInfo) setReqHeaders(req *http.Request) {
 }
 
 func (a *AccessInfo) CheckAccess(request *authzv1.SubjectAccessReviewSpec) (*authzv1.SubjectAccessReviewStatus, error) {
+	klog.V(5).Infof("CheckAccess request for user %s, resource attributes: %s", request.User, request.ResourceAttributes)
```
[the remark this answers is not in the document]
artunduman: no it's not, but I think it was showing up okay in the logs. Do you prefer I change it to %v?

authz/providers/azure/rbac/rbac.go
artunduman: There shouldn't be, because of L386

authz/providers/azure/rbac/rbac.go
artunduman: I think this one can stay as Infof, since this is a common occurrence and not related to any component health

authz/providers/azure/rbac/rbac.go
artunduman: Because this code is on a really critical path, we didn't want to do anything that can potentially cause any regression (or a security incident) for the existing behavior. That's why I didn't extract/refactor.

authz/providers/azure/rbac/rbac.go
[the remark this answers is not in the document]
saisankargochhayat: Might be nice to have `%+v`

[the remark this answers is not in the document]
artunduman: Actually it has a default .String() attribute (I think it comes from protobuf). So I guess I can keep it as is if there's no objections
`func (this *ResourceAttributes) String() string {
	if this == nil {
		return "nil"
	}
	s := strings.Join([]string{`&ResourceAttributes{`,
		`Namespace:` + fmt.Sprintf("%v", this.Namespace) + `,`,
		`Verb:` + fmt.Sprintf("%v", this.Verb) + `,`,
		`Group:` + fmt.Sprintf("%v", this.Group) + `,`,
		`Version:` + fmt.Sprintf("%v", this.Version) + `,`,
		`Resource:` + fmt.Sprintf("%v", this.Resource) + `,`,
		`Subresource:` + fmt.Sprintf("%v", this.Subresource) + `,`,
		`Name:` + fmt.Sprintf("%v", this.Name) + `,`,
		`}`,
	}, "")
	return s
}`

authz/providers/azure/rbac/rbac.go
```diff
@@ -375,7 +378,94 @@ func (a *AccessInfo) CheckAccess(request *authzv1.SubjectAccessReviewSpec) (*aut
 
 		finalStatus = status
 	}
-	return finalStatus, nil
+	if finalStatus != nil && finalStatus.Allowed {
+		klog.V(5).Infof("Checkaccess request is allowed for user %s", checkAccessUsername)
+		return finalStatus, nil
+	}
+
+	if !a.useManagedNamespaceResourceScopeFormat || a.clusterType != managedClusters {
```
saisankargochhayat, line 386: Since we already know this is it necessary to traverse both paths?
`Logically, it means Guard will allow the request iff (#1 OR #2a OR #2b). `
Can we not do either #1 or (#2a or #2b)?

authz/providers/azure/rbac/rbac.go
```diff
@@ -375,7 +378,94 @@ func (a *AccessInfo) CheckAccess(request *authzv1.SubjectAccessReviewSpec) (*aut
...
+
+	klog.V(5).Infof("Falling back to checking managed namespace scope", checkAccessUsername)
+
+	checkAccessURLManagedNS := *a.apiURL
+	checkAccessURLManagedNS.Path = path.Join(checkAccessURLManagedNS.Path, a.azureResourceId)
+
+	exists, managedNameSpaceString := getManagedNameSpaceScope(request)
+	if !exists {
+		// cluster level request
+		klog.V(5).Infof("Skipping managed namespace check for user %s because SAR is cluster scoped", checkAccessUsername)
```
saisankargochhayat, line 399: nit: SAR -> subject access review

authz/providers/azure/rbac/rbac.go
artunduman: Hmm can you elaborate on that? The 2b is not implemented yet, that's for a later step of the project

saisankargochhayat: I mean does it have to be a fallback or based on `a.useManagedNamespaceResourceScopeFormat` can we switch between 1 and 2a?

authz/providers/azure/rbac/rbac.go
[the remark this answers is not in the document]
artunduman: Yeah I realized that - When I was debugging, I couldn't find anything in the logs to show what the SAR is actually requesting, I could only see whether denied or allowed. That's why I wanted to add.
I can make it more than 5 if you think this would spam, but it only contains a few fields that show the NS, group, verb etc. Not the whole object that is being requested.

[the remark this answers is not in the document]
artunduman: Let me know if that's what you prefer and I can make it klog.V(7) or something :)

weinong:
Can you add more detail in the PR description on why we implement it this way? the logical OR is a fact, but no one would understand the why.

authz/providers/azure/rbac/rbac.go
weinong: can we refactor them into function? my concern is if we will change the code in one place and miss in the other.

weinong: the code itself is already complicated. duplicating that will make people scratch their head even harder

artunduman: refactored

authz/providers/azure/rbac/rbac.go
```diff
@@ -378,6 +364,65 @@ func (a *AccessInfo) CheckAccess(request *authzv1.SubjectAccessReviewSpec) (*aut
...
+	paramsManagedNS.Add("api-version", checkAccessAPIVersion)
+	managedNamespaceURL.RawQuery = paramsManagedNS.Encode()
+
+	// update resource IDs for managed namespace
+	for _, b := range checkAccessBodies {
+		b.Resource.Id = path.Join(a.azureResourceId, managedNamespacePath)
+	}
+
+	return a.performCheckAccess(managedNamespaceURL, checkAccessBodies, checkAccessUsername)
+}
```
weinong, line 424: I used copilot to refactor this code to make it less repetitive. WDYT?

```go
func (a *AccessInfo) CheckAccess(request *authzv1.SubjectAccessReviewSpec) (*authzv1.SubjectAccessReviewStatus, error) {
checkAccessBodies, err := prepareCheckAccessRequestBody(request, a.clusterType, a.azureResourceId, a.useNamespaceResourceScopeFormat)
if err != nil {
return nil, errors.Wrap(err, "error in preparing check access request")
}

checkAccessUsername := request.User

// Build primary check access URL
exist, nameSpaceString := getNameSpaceScope(request, a.useNamespaceResourceScopeFormat)
checkAccessURL := buildCheckAccessURL(*a.apiURL, a.azureResourceId, exist, nameSpaceString)

status, err := a.performCheckAccess(checkAccessURL, checkAccessBodies, checkAccessUsername)
if err != nil {
return nil, err
}
if status != nil && status.Allowed {
klog.V(7).Infof("Checkaccess request is allowed for user %s", checkAccessUsername)
return status, nil
}

// Fallback to managed namespace check
if !a.useManagedNamespaceResourceScopeFormat || a.clusterType != managedClusters {
klog.V(7).Infof("Checkaccess request is denied for user %s", checkAccessUsername)
return status, nil
}

klog.V(7).Infof("Falling back to checking managed namespace scope for user %s", checkAccessUsername)
exists, managedNamespacePath := getManagedNameSpaceScope(request)
if !exists {
klog.V(7).Infof(
"Skipping managed namespace check for user %s because subject access review is cluster scoped",
checkAccessUsername,
)
return status, nil
}

// Build managed namespace URL
managedNamespaceURL := buildCheckAccessURL(*a.apiURL, a.azureResourceId, true, managedNamespacePath)

// Update resource IDs for managed namespace
for _, b := range checkAccessBodies {
b.Resource.Id = path.Join(a.azureResourceId, managedNamespacePath)
}

return a.performCheckAccess(managedNamespaceURL, checkAccessBodies, checkAccessUsername)
}

// Helper to build the check access URL
func buildCheckAccessURL(base url.URL, resourceID string, hasNamespace bool, namespacePath string) url.URL {
base.Path = path.Join(base.Path, resourceID)
if hasNamespace {
base.Path = path.Join(base.Path, namespacePath)
}
base.Path = path.Join(base.Path, checkAccessPath)

params := url.Values{}
params.Add("api-version", checkAccessAPIVersion)
base.RawQuery = params.Encode()

return base
}
```

artunduman: Looks good to me, updating the PR with this

artunduman: updated

weinong (approved):
LGTM

PR #414, opened 2025-06-05
Support tracked managed namespace deletions

artunduman:


authz/providers/azure/rbac/rbac.go
```diff
@@ -427,7 +428,7 @@ func buildCheckAccessURL(base url.URL, resourceID string, hasNamespace bool, nam
 	return base
 }
 
-func (a *AccessInfo) sendCheckAccessRequest(ctx context.Context, checkAccessUsername string, checkAccessURL url.URL, checkAccessBody *CheckAccessRequest, ch chan *authzv1.SubjectAccessReviewStatus) error {
+func (a *AccessInfo) sendCheckAccessRequest(ctx context.Context, checkAccessUsername string, checkAccessURL url.URL, checkAccessBody *CheckAccessRequest, ch chan *authzv1.SubjectAccessReviewStatus, isTrackedResource bool) error {
```
chengliangli0918, line 431: do we really need this argument 'isTrackedResource bool'? 

For a non-tracked child resource, regardless its existence,  'az role assignment list ...' will return an empty array of role assignment array if there is any. 

For a tracked child resource that ARM knows of,  'az role assignment list ...' will return 'ResourceNotFound' if this tracked child resource doesn't exist, and return a role assignment array (if there is any) for this resource if it exists.

If you have this additional, when we switch the routing type of managedClusters/managedNamespaces between 'proxyonly' and 'tracked', we always need to update the guard logic here.
image

authz/providers/azure/rbac/rbac.go
```diff
@@ -504,10 +506,16 @@ func (a *AccessInfo) sendCheckAccessRequest(ctx context.Context, checkAccessUser
 		checkAccessSucceeded.Inc()
 	}
 
-	// Decode response and prepare k8s response
-	status, err := ConvertCheckAccessResponse(checkAccessUsername, data)
-	if err != nil {
-		return err
+	var status *authzv1.SubjectAccessReviewStatus
+	if isTrackedResource && resp.StatusCode == http.StatusNotFound {
```
chengliangli0918, line 510: According to my experiments, StatusNotFound can only be returned for Tracked child resources. Top level resource cannot be 'ProxyOnly'. so isTrackedResource is not needed. See my previous comment for more details.

authz/providers/azure/rbac/rbac.go
```diff
@@ -517,7 +517,8 @@ func (a *AccessInfo) sendCheckAccessRequest(ctx context.Context, checkAccessUser
 	}
 
 	klog.V(7).Infof("checkaccess response: %s, Configured ARM call limit: %d", string(data), a.armCallLimit)
-	if resp.StatusCode != http.StatusOK {
+	// We can expect the response to be a 200 OK for proxy resources or 404 Not Found for non-proxy resources due to resource deletion
```
weinong, line 520:
```suggestion
	// We can expect the response to be a 200 OK for ARM proxy resources or 404 Not Found for ARM tracked resources due to resource deletion
```
i suggest using ARM terminology here as "proxy" can be interpreted under different context.

authz/providers/azure/rbac/rbac.go
```diff
@@ -547,10 +548,16 @@ func (a *AccessInfo) sendCheckAccessRequest(ctx context.Context, checkAccessUser
 		checkAccessSucceeded.Inc()
 	}
 
-	// Decode response and prepare k8s response
-	status, err := ConvertCheckAccessResponse(checkAccessUsername, data)
-	if err != nil {
-		return err
+	var status *authzv1.SubjectAccessReviewStatus
```
weinong, line 551: the else in L535 is redundant. removing that will improve readability.

weinong (approved):
LGTM

PR #424, opened 2025-10-16
Context aware logging and customResourceCheckCoverate for azure rbac.

arxhive:
Added context-aware logging including request ID tracking that glue authorization request logging for azure rbac:

  - authz/providers/azure/azure.go:
    - Added a unique request ID (UUID) generated at the start of each authorization check
    - Request ID flows through the entire authorization workflow
    - All log statements now include the request ID for easy filtering
  - authz/providers/azure/rbac/rbac.go:
    - Request ID is extracted from context and passed to all sub-functions
    - CheckAccess, performCheckAccess, and sendCheckAccessRequest all log with request ID
    - Each batch request also gets a correlation ID that's logged alongside the parent request ID

  2. Enhanced Throttling Detection and Logging

  - Explicit HTTP 429 detection:
    - Clear log message: "Azure throttling detected (HTTP 429), closing idle TCP connections to switch ARM instance"
    - Includes request ID, correlation ID, and user information
  - ARM call limit threshold logging:
    - Logs when remaining ARM calls drop below threshold
    - Shows: "ARM call limit threshold reached, closing idle TCP connections to switch ARM instance"
    - Includes remaining call count and configured threshold

  3. Retry Attempt Logging with Request Context

  - Custom context-aware logger:
    - Created a custom logger for the retry HTTP client
    - Automatically extracts and logs request ID and correlation ID from context
    - All retry attempts now include request context in logs

  4. Structured Logging Throughout

  All logs now use klog.InfoS, klog.ErrorS, and klog.V().InfoS with structured key-value pairs:
  - Request lifecycle:
    - Start: "Authorization check started" with requestID
    - End: "Authorization check completed" or "Authorization check failed" with full context
  - CheckAccess operations:
    - Sending: "Sending checkAccess request to Azure"
    - Success: "CheckAccess request succeeded" with remainingARMCalls, durationSeconds
    - Failure: "CheckAccess request failed" with statusCode, response body, duration

  5. Added comprehensive logging for the AllowCustomResourceTypeCheck feature

### DIFF
--- a/authz/providers/azure/rbac/checkaccessreqhelper.go
+++ b/authz/providers/azure/rbac/checkaccessreqhelper.go
@@ -265,9 +265,9 @@ func getDataActions(ctx context.Context, subRevReq *authzv1.SubjectAccessReviewS
 	if subRevReq.ResourceAttributes != nil {
 		storedOperationsMap := getStoredOperationsMap()
 
-		isCustomerResourceTypeCheckAvailable := allowCustomResourceTypeCheck && storedOperationsMap != nil && len(storedOperationsMap) != 0
+		isCustomerResourceTypeCheckAvailable := allowCustomResourceTypeCheck && len(storedOperationsMap) != 0
 		if !isCustomerResourceTypeCheckAvailable {
-			log.V(5).Info("CustomResourceTypeCheck feature is not available", "allowCustomResourceTypeCheck", allowCustomResourceTypeCheck, "operationsMapAvailable", storedOperationsMap != nil && len(storedOperationsMap) != 0)
+			log.V(5).Info("CustomResourceTypeCheck feature is not available", "allowCustomResourceTypeCheck", allowCustomResourceTypeCheck, "operationsMapAvailable", len(storedOperationsMap) != 0)
 		}
 
 		if subRevReq.ResourceAttributes.Resource != "*" && subRevReq.ResourceAttributes.Group != "*" && subRevReq.ResourceAttributes.Verb != "*" {
@@ -309,7 +309,7 @@ func getDataActions(ctx context.Context, subRevReq *authzv1.SubjectAccessReviewS
 			authInfoList = append(authInfoList, authInfoSingle)
 
 		} else {
-			if storedOperationsMap == nil || (storedOperationsMap != nil && len(storedOperationsMap) == 0) {
+			if len(storedOperationsMap) == 0 {
 				return nil, fmt.Errorf("Wildcard support for Resource/Verb/Group is not enabled for request Group: %s, Resource: %s, Verb: %s", subRevReq.ResourceAttributes.Group, subRevReq.ResourceAttributes.Resource, subRevReq.ResourceAttributes.Verb)
 			}
 


authz/providers/azure/rbac/checkaccessreqhelper.go
```diff
@@ -250,7 +250,7 @@ func getDataActions(subRevReq *authzv1.SubjectAccessReviewSpec, clusterType stri
 
 		isCustomerResourceTypeCheckAvailable := allowCustomResourceTypeCheck && storedOperationsMap != nil && len(storedOperationsMap) != 0
 		if !isCustomerResourceTypeCheckAvailable {
-			klog.V(5).Info("CustomResource type verification is not available for this request")
+			klog.V(5).Info("CustomResourceTypeCheck feature is not available for this request", "allowCustomResourceTypeCheck", allowCustomResourceTypeCheck, "operationsMapAvailable", storedOperationsMap != nil && len(storedOperationsMap) != 0)
```
weinong: here is why i like https://github.com/go-logr/logr more. you can create the logger with value at the beginning. then you don't need to repeat `CustomResourceTypeCheck` or `CustomResourceTypeCheck:` all over in this function

authz/providers/azure/rbac/checkaccessreqhelper.go
```diff
@@ -429,25 +432,30 @@ func getAuthInfoListForWildcard(subRevReq *authzv1.SubjectAccessReviewSpec, stor
...
 
 	for i := range authInfoList {
 		err := setAuthInfoResourceAttributes(&authInfoList[i], subRevReq)
 		if err != nil {
+			klog.ErrorS(err, "CustomResourceTypeCheck: Error while setting resource attributes", "group", subRevReq.ResourceAttributes.Group, "resource", subRevReq.ResourceAttributes.Resource)
 			return nil, errors.Errorf("Error while setting resource attributes: %s", err.Error())
 		}
 	}
 
+	klog.V(5).InfoS("CustomResourceTypeCheck: Successfully created authorization actions for custom resource", "group", subRevReq.ResourceAttributes.Group, "resource", subRevReq.ResourceAttributes.Resource, "actionsCount", len(authInfoList))
```
weinong: do we really want to log when it's successful? that'd be a lot

authz/providers/azure/rbac/checkaccessreqhelper.go
```diff
@@ -429,25 +432,30 @@ func getAuthInfoListForWildcard(subRevReq *authzv1.SubjectAccessReviewSpec, stor
...
 		}
+		klog.V(7).InfoS("CustomResourceTypeCheck: Creating action for custom resource", "group", subRevReq.ResourceAttributes.Group, "resource", subRevReq.ResourceAttributes.Resource, "verb", subRevReq.ResourceAttributes.Verb, "action", action)
 		authInfoList = append(authInfoList, authInfoSingle)
 	}
 
 	for i := range authInfoList {
 		err := setAuthInfoResourceAttributes(&authInfoList[i], subRevReq)
 		if err != nil {
+			klog.ErrorS(err, "CustomResourceTypeCheck: Error while setting resource attributes", "group", subRevReq.ResourceAttributes.Group, "resource", subRevReq.ResourceAttributes.Resource)
 			return nil, errors.Errorf("Error while setting resource attributes: %s", err.Error())
```
weinong: where is this error observed/logged? if it's already logged, i suggest removing the log above.

authz/providers/azure/rbac/checkaccessreqhelper.go
arxhive: removed

authz/providers/azure/rbac/checkaccessreqhelper.go
```diff
@@ -271,6 +271,7 @@ func getDataActions(subRevReq *authzv1.SubjectAccessReviewSpec, clusterType stri
 					In this case both Res and Group are not *, but there is no matching DataAction present on the storedOperationsMap.
 					The resource is presumed to be a CR and <clusterType>/customresources/<action> DataAction will be used for check access.
 				*/
+				klog.V(5).InfoS("CustomResourceTypeCheck: Resource identified as custom resource", "group", subRevReq.ResourceAttributes.Group, "resource", subRevReq.ResourceAttributes.Resource, "verb", subRevReq.ResourceAttributes.Verb)
```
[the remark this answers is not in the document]
arxhive: What about keeping ResourceAttributes.Resource in log context?

authz/providers/azure/azure.go
```diff
@@ -84,32 +85,42 @@ func newAuthzClient(opts authzOpts.Options, authopts auth.Options) (authz.Interf
...
+
+	log := klog.FromContext(ctx).WithValues("requestID", requestID, "resourceAttributes", request.ResourceAttributes)
+	ctx = klog.NewContext(ctx, log)
+
 	if request == nil {
-		return nil, errutils.WithCode(errors.New("subject access review is nil"), http.StatusBadRequest)
+		return nil, errutils.WithCode(errors.Errorf("Authorization request failed (requestID: %s): subject access review is nil", requestID), http.StatusBadRequest)
 	}
 
+	log.Info("Authorization check started")
```
weinong: wouldn't this be too noisy

authz/providers/azure/azure.go
```diff
@@ -84,32 +85,42 @@ func newAuthzClient(opts authzOpts.Options, authopts auth.Options) (authz.Interf
...
 		} else {
+			log.Info("Non-AAD user denied")
 			return &authzv1.SubjectAccessReviewStatus{Allowed: false, Denied: true, Reason: rbac.NonAADUserNotAllowedVerdict}, nil
 		}
 	}
 
 	exist, result := s.rbacClient.GetResultFromCache(request, store)
 
 	if exist {
+		log.V(5).Info("Cache hit", "allowed", result)
```
weinong: this is where we would want to add resource attributes.

authz/providers/azure/azure.go
```diff
@@ -119,29 +130,32 @@ func (s Authorizer) Check(ctx context.Context, request *authzv1.SubjectAccessRev
...
 			return nil, errutils.WithCode(err, http.StatusInternalServerError)
 		}
+		log.V(5).Info("Token refreshed successfully")
 	}
 
-	response, err := s.rbacClient.CheckAccess(request)
+	response, err := s.rbacClient.CheckAccess(ctx, request)
 	if err == nil {
-		klog.V(5).Infof(response.Reason)
+		log.Info("Authorization check completed", "allowed", response.Allowed, "reason", response.Reason)
```
weinong: this is where we would want to log resource attributes

authz/providers/azure/rbac/rbac.go
```diff
@@ -232,8 +229,7 @@ func (a *AccessInfo) RefreshToken(ctx context.Context) error {
 	if a.IsTokenExpired() {
 		resp, err := a.tokenProvider.Acquire(ctx, "")
 		if err != nil {
-			klog.Errorf("%s failed to refresh token : %s", a.tokenProvider.Name(), err.Error())
-			return errors.Wrap(err, "failed to refresh rbac token")
+			return errors.Wrapf(err, "Failed to refresh token (provider: %s)", a.tokenProvider.Name())
```
weinong: it doesn't capture the request ID, does it?

weinong:
general feedback
1. please remove the resource attributes from the context logger for being too verbose. Also, it doesn't capture for non resources.
2. please look into how and where to log error with the requestID
3. can you update the log [here](https://github.com/kubeguard/guard/blob/master/authz/providers/azure/rbac/checkaccessreqhelper.go#L310)? it doesn't tell which condition which is hard to troubleshoot.
4. while logging cache hit or miss is useful, we still need insights into the cache itself. please look into https://github.com/kubeguard/guard/blob/master/authz/providers/azure/data/datastore.go.
5. regarding exec data action we discussed last time, we should probably log the role definition ID returned by checkAccess api? that'd help us identify if `exec` is authorized based on any built-in roles or not.

util/azure/utils.go
```diff
@@ -506,9 +505,19 @@ func fetchDataActionsList(ctx context.Context) ([]Operation, error) {
...
+
+func (l *contextAwareLogger) Printf(format string, v ...interface{}) {
+	// Use klog's context-aware logging which automatically includes request ID and correlation ID
+	klog.FromContext(l.ctx).Info(fmt.Sprintf(format, v...))
+}
+
 // MakeRetryableHttpClient creates an HTTP client which attempts the request
 // (1 + retryCount) times and has a 3 second timeout per attempt.
-func MakeRetryableHttpClient(retryCount int) retryablehttp.Client {
+func MakeRetryableHttpClient(retryCount int, ctx context.Context) retryablehttp.Client {
```
karataliu: making context 1st parameter, following convention?

authz/providers/azure/rbac/checkaccessreqhelper.go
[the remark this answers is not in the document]
arxhive: Agreed! Resource attributes are now only logged at key decision points (cache hit, authorization completed) with requestID providing correlation across all logs.

authz/providers/azure/rbac/checkaccessreqhelper.go
```diff
@@ -429,25 +432,30 @@ func getAuthInfoListForWildcard(subRevReq *authzv1.SubjectAccessReviewSpec, stor
...
+		klog.V(7).InfoS("CustomResourceTypeCheck: Creating actions for all custom resource verbs", "group", subRevReq.ResourceAttributes.Group, "resource", subRevReq.ResourceAttributes.Resource)
 		for _, action := range getCustomResourceOperationsMap(clusterType) {
 			authInfoList = append(authInfoList, action)
 		}
 	} else {
 		action := getActionName(subRevReq.ResourceAttributes.Verb)
 		authInfoSingle, found := getCustomResourceOperationsMap(clusterType)[action]
 		if !found {
+			klog.ErrorS(errors.New("action not found"), "CustomResourceTypeCheck: No actions found for custom resource verb", "verb", subRevReq.ResourceAttributes.Verb, "group", subRevReq.ResourceAttributes.Group, "resource", subRevReq.ResourceAttributes.Resource)
 			return nil, errors.Errorf("No actions found for verb %s", subRevReq.ResourceAttributes.Verb)
```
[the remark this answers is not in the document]
arxhive: Added the action to the returned error message. Removed the ErrorS log to avoid duplicate logging - the error is returned and logged at the call site with full context.

authz/providers/azure/azure.go
arxhive: Agreed and removed this log to reduce noise.

authz/providers/azure/azure.go
```diff
@@ -84,32 +85,42 @@ func newAuthzClient(opts authzOpts.Options, authopts auth.Options) (authz.Interf
 }
 
 func (s Authorizer) Check(ctx context.Context, request *authzv1.SubjectAccessReviewSpec, store authz.Store) (*authzv1.SubjectAccessReviewStatus, error) {
+	requestID := uuid.New().String()
+
+	log := klog.FromContext(ctx).WithValues("requestID", requestID, "resourceAttributes", request.ResourceAttributes)
```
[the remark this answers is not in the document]
arxhive: Removed resourceAttributes from the global log context. Since we have requestID for correlation, resource attributes are now only logged at specific important points (cache hit and authorization completed) using InfoS for proper structured logging. This reduces verbosity while maintaining debuggability.

authz/providers/azure/azure.go
arxhive: Added resourceAttributes to this log using InfoS for structured logging. This is one of the key decision points where we want to see the full context.

authz/providers/azure/rbac/rbac.go
```diff
@@ -464,71 +450,81 @@ func (a *AccessInfo) CheckAccess(request *authzv1.SubjectAccessReviewSpec) (*aut
...
 	duration := time.Since(start).Seconds()
 	if err != nil {
 		checkAccessTotal.WithLabelValues(internalServerCode).Inc()
 		checkAccessDuration.WithLabelValues(internalServerCode).Observe(duration)
-		return errutils.WithCode(errors.Wrap(err, "error in check access request execution."), http.StatusInternalServerError)
+		return errutils.WithCode(errors.Wrapf(err, "CheckAccess request execution failed (durationSeconds: %.2f)", duration), http.StatusInternalServerError)
 	}
 
-	defer resp.Body.Close()
+	defer func() {
```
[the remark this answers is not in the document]
arxhive: Good question! The wrapper syntax `defer func() { _ = resp.Body.Close() }()` is preferred over simple `defer resp.Body.Close()` because:

1. **Explicit error handling** - `resp.Body.Close()` returns an error. The wrapper with `_ =` explicitly shows we're intentionally ignoring this error, making the intent clear to code reviewers and linters.

2. **Linter compliance** - Some linters (like `errcheck`) will flag unchecked errors. The `_ =` assignment satisfies these linters by showing the error is deliberately ignored.

3. **Code clarity** - It signals to future maintainers that we considered the error and decided it's safe to ignore in this defer context (since we're already handling the main response processing errors elsewhere).

The simple `defer resp.Body.Close()` works the same way functionally, but doesn't communicate this intent as clearly.

authz/providers/azure/rbac/rbac.go
```diff
@@ -464,71 +450,81 @@ func (a *AccessInfo) CheckAccess(request *authzv1.SubjectAccessReviewSpec) (*aut
...
+		return errutils.WithCode(errors.Wrap(err, "Failed to create check access request"), http.StatusInternalServerError)
 	}
 
 	a.setReqHeaders(req)
-	// set x-ms-correlation-request-id for the checkaccess request
-	correlationID := ctx.Value(correlationRequestIDKey(correlationRequestIDHeader)).([]string)
-	req.Header[correlationRequestIDHeader] = correlationID
+
+	// Get logger from context (already has correlationID and batchIndex from parent)
+	log = klog.FromContext(ctx)
```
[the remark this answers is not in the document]
arxhive: Yes, this is intentional. Line 496 gets the logger from context (which includes the parent requestID). Line 510 gets it again because we're inside a goroutine and need to ensure we have the current context's logger with the correlation ID and batch index that were added at line 332.

authz/providers/azure/rbac/checkaccessreqhelper.go
```diff
@@ -256,16 +257,20 @@ func getResourceAndAction(resource string, subResource string, verb string) stri
...
-func getDataActions(subRevReq *authzv1.SubjectAccessReviewSpec, clusterType string, allowCustomResourceTypeCheck bool, allowSubresourceTypeCheck bool) ([]azureutils.AuthorizationActionInfo, error) {
+func getDataActions(ctx context.Context, subRevReq *authzv1.SubjectAccessReviewSpec, clusterType string, allowCustomResourceTypeCheck bool, allowSubresourceTypeCheck bool) ([]azureutils.AuthorizationActionInfo, error) {
 	var authInfoList []azureutils.AuthorizationActionInfo
 	var err error
 
 	if subRevReq.ResourceAttributes != nil {
 		storedOperationsMap := getStoredOperationsMap()
 
+		// Create logger with feature context for custom resource type check
+		featureLog := klog.FromContext(ctx).WithValues("feature", "CustomResourceTypeCheck")
```
weinong: i don't quite follow how this constant tag would be usefu?

authz/providers/azure/azure.go
```diff
@@ -119,29 +128,31 @@ func (s Authorizer) Check(ctx context.Context, request *authzv1.SubjectAccessRev
 
 	// if set true, webhook will allow access to discovery APIs for authenticated users. If false, access check will be performed on Azure.
 	if s.rbacClient.AllowNonResPathDiscoveryAccess(request) {
-		klog.V(10).Infof("Allowing user %s access for discovery check.", request.User)
-		_ = s.rbacClient.SetResultInCache(request, true, store)
+		log.V(10).Info("Allowing user access for discovery check")
+		_ = s.rbacClient.SetResultInCache(ctx, request, true, store)
```
weinong: we should log the error here so that we don't need to put logging in the cache library

authz/providers/azure/data/datastore.go
```diff
@@ -41,55 +43,99 @@ type DataStore struct {
 
 // Set stores the given value for the given key.
 // The key must not be "" and the value must not be nil.
-func (s *DataStore) Set(key string, value interface{}) error {
+func (s *DataStore) Set(ctx context.Context, key string, value interface{}) error {
+	log := klog.FromContext(ctx)
```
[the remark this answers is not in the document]
arxhive: Alright, rolled back datastore.go. But also enriched errors with cache statistics. The caller (rbac.go) logs cache errors with full requestID context where appropriate.

authz/providers/azure/azure.go
arxhive: Updated to 
```go
if err := s.rbacClient.SetResultInCache(ctx, request, true, store); err != nil {
			log.Error(err, "Failed to cache discovery access result")
		}
```

authz/providers/azure/rbac/checkaccessreqhelper.go
```diff
@@ -673,7 +684,7 @@ func prepareCheckAccessRequestBody(req *authzv1.SubjectAccessReviewSpec, cluster
 		return nil, errutils.WithCode(errors.New("oid info not sent from authentication module"), http.StatusBadRequest)
```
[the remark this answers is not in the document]
arxhive: Good catch, attached  requestId to the error
```go
			return nil, errutils.WithCode(fmt.Errorf("oid info sent from authentication module is not valid (requestID: %s, oid: %s)", getRequestID(ctx), userOid), http.StatusBadRequest)
```

authz/providers/azure/azure.go
```diff
@@ -119,29 +128,31 @@ func (s Authorizer) Check(ctx context.Context, request *authzv1.SubjectAccessRev
...
-		_ = s.rbacClient.SetResultInCache(request, response.Allowed, store)
+		log.InfoS("Authorization check completed", "allowed", response.Allowed, "reason", response.Reason, "resourceAttributes", request.ResourceAttributes)
+		_ = s.rbacClient.SetResultInCache(ctx, request, response.Allowed, store)
 	} else {
 		code := http.StatusInternalServerError
 		if v, ok := err.(errutils.HttpStatusCode); ok {
 			code = v.Code()
 		}
-		err = errutils.WithCode(errors.Errorf(rbac.CheckAccessErrorFormat, err), code)
+		// Error already contains requestID from context, just add status code
```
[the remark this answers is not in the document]
arxhive: Attached requestID
```go
			return nil, errutils.WithCode(errors.Wrapf(err, "Failed to refresh token (requestID: %s)", requestID), http.StatusInternalServerError)

```

authz/providers/azure/rbac/checkaccessreqhelper.go
```diff
@@ -667,13 +687,13 @@ func prepareCheckAccessRequestBody(req *authzv1.SubjectAccessReviewSpec, cluster
 		val := oid.String()
 		userOid = val[1 : len(val)-1]
 		if !isValidUUID(userOid) {
-			return nil, errutils.WithCode(errors.New("oid info sent from authentication module is not valid"), http.StatusBadRequest)
+			return nil, errutils.WithCode(fmt.Errorf("oid info sent from authentication module is not valid (requestID: %s, oid: %s)", getRequestID(ctx), userOid), http.StatusBadRequest)
```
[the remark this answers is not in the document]
weinong: for instance, you missed (?) including it at L693. So, including requestID at this level is very not reliable. The best place is as you did when the error is observed at the highest level we cantrol

authz/providers/azure/rbac/checkaccessreqhelper.go
```diff
@@ -34,6 +35,17 @@ import (
...
+type contextKey string
+
+const requestIDKey contextKey = "requestID"
+
+func getRequestID(ctx context.Context) string {
+	if requestID, ok := ctx.Value(requestIDKey).(string); ok {
+		return requestID
+	}
+	return "unknown"
+}
```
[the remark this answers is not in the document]
arxhive: Good idea!

authz/providers/azure/rbac/checkaccessreqhelper.go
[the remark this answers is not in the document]
arxhive: Removed requestIds from checkaccessreqhelper.go

weinong (commented):
`github.com/pkg/errors` has been archived. i wonder if you can update it as well?

authz/providers/azure/rbac/checkaccessreqhelper.go
```diff
@@ -736,6 +745,13 @@ func ConvertCheckAccessResponse(username string, body []byte) (*authzv1.SubjectA
 	if deniedResultFound == -1 { // no denied result found
 		allowed = true
 		verdict = fmt.Sprintf(AccessAllowedVerboseVerdict, response[0].AzureRoleAssignment.Id, response[0].AzureRoleAssignment.RoleDefinitionId, username)
+
+		// Log role definition ID to help identify if exec/other actions are authorized via built-in or custom roles
+		log.V(7).Info("Access allowed via role assignment",
```
weinong: to allow us to evaluate it, we will need to change the verbose level to 7, which will inherently bring in more logs. i think changing it to 5 is better in terms of managing log volume

authz/providers/azure/data/datastore.go
```diff
@@ -18,12 +18,13 @@ package data
...
 import (
 	"encoding/json"
-	"errors"
+	"fmt"
 	"time"
 
 	"go.kubeguard.dev/guard/authz"
 
 	"github.com/allegro/bigcache"
+	"github.com/pkg/errors"
```
weinong: i'd rather stick with `errors` package which is more modern

arxhive:
> `github.com/pkg/errors` has been archived. i wonder if you can update it as well?

yeah, good idea, migrated to a native errors package in this commit :)
https://github.com/kubeguard/guard/pull/424/commits/a2741a77320fe0a5f623772f07488717e397ed5b

authz/providers/azure/rbac/rbac.go
```diff
@@ -256,18 +254,28 @@ func (a *AccessInfo) ShouldSkipAuthzCheckForNonAADUsers() bool {
...
 	var result bool
 	key := getResultCacheKey(request, a.allowSubresourceTypeCheck)
-	klog.V(10).Infof("Cache search for key: %s", key)
-	found, _ := store.Get(key, &result)
+	log.V(10).Info("Cache search", "key", key)
+	found, err := store.Get(key, &result)
+
+	if err != nil {
+		// Error contains cache statistics for troubleshooting
+		log.V(8).Info("Cache get error", "key", key, "error", err)
```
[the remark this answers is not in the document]
arxhive: 5 :)

authz/providers/azure/rbac/checkaccessreqhelper.go
```diff
@@ -256,16 +257,17 @@ func getResourceAndAction(resource string, subResource string, verb string) stri
...
 }
 
-func getDataActions(subRevReq *authzv1.SubjectAccessReviewSpec, clusterType string, allowCustomResourceTypeCheck bool, allowSubresourceTypeCheck bool) ([]azureutils.AuthorizationActionInfo, error) {
+func getDataActions(ctx context.Context, subRevReq *authzv1.SubjectAccessReviewSpec, clusterType string, allowCustomResourceTypeCheck bool, allowSubresourceTypeCheck bool) ([]azureutils.AuthorizationActionInfo, error) {
 	var authInfoList []azureutils.AuthorizationActionInfo
 	var err error
 
 	if subRevReq.ResourceAttributes != nil {
 		storedOperationsMap := getStoredOperationsMap()
+		log := klog.FromContext(ctx)
```
weinong: nit: it should be around L262

authz/providers/azure/rbac/rbac.go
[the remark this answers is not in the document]
arxhive: pushed a few mins ago, this and a few other logs

authz/providers/azure/rbac/checkaccessreqhelper.go
arxhive: 5 👍

authz/providers/azure/azure.go
```diff
@@ -119,29 +129,33 @@ func (s Authorizer) Check(ctx context.Context, request *authzv1.SubjectAccessRev
...
+		if err := s.rbacClient.SetResultInCache(ctx, request, response.Allowed, store); err != nil {
+			log.Error(err, "Failed to cache authorization result")
+		}
 	} else {
 		code := http.StatusInternalServerError
 		if v, ok := err.(errutils.HttpStatusCode); ok {
 			code = v.Code()
 		}
-		err = errutils.WithCode(errors.Errorf(rbac.CheckAccessErrorFormat, err), code)
+		err = errutils.WithCode(fmt.Errorf("Authorization check failed (requestID: %s, statusCode: %d): %w", requestID, code, err), code)
```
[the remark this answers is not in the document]
arxhive: Don't we trust this thing?
`authzhandler.go -> writeAuthzResponse(w, &data.Spec, resp, err) -> printStackTrace`

weinong (approved):
lgtm

PR #426, opened 2025-11-14
Migration to checkaccess v2 migration for azure authz provider

arxhive:
This PR implements support for Azure's CheckAccess v2 API while maintaining full backward compatibility with the existing v1 implementation. The v2 API uses the official Azure SDK (github.com/Azure/checkaccess-v2-go-sdk) and can be enabled via feature flags.

  Changes

  1. CheckAccess v2 SDK Integration (authz/providers/azure/rbac/checkaccess_v2.go)
    - Official Azure SDK integration for RBAC authorization checks
    - Maintains same batching behavior as v1 (200 actions per request)
    - Multi-level fallback: primary → managed namespace → fleet
    - Prometheus metrics consistent with v1 patterns
  2. Token Credential Adapter (authz/providers/azure/rbac/tokencredential_adapter.go)
    - Adapts existing graph.TokenProvider to Azure SDK's azcore.TokenCredential
    - Enables reuse of all existing token providers (AKS, ARC, client credentials, MSI)
  3. Configuration Options (authz/providers/azure/options/options.go)
    - --azure.use-checkaccess-v2: Enable v2 API (default: false)
    - --azure.pdp-endpoint: PDP endpoint URL (required when v2 enabled)
    - --azure.checkaccess-v2-api-version: Optional API version override
  4. Dual-mode Support (authz/providers/azure/rbac/rbac.go)
    - Routes requests to v2 or v1 based on configuration
    - v1 remains the default, ensuring backward compatibility

  
  Migration Path

  1. Phase 1 (this PR): Deploy with v1 as default (no changes needed)
  2. Phase 2: Enable v2 in staging with --azure.use-checkaccess-v2=true
  3. Phase 3: Roll out v2 to production clusters progressively
  4. Future: Deprecate v1 after successful v2 adoption

### DIFF
--- a/authz/providers/azure/rbac/rbac.go
+++ b/authz/providers/azure/rbac/rbac.go
@@ -86,7 +86,6 @@ type AccessInfo struct {
 	// These fields use the official Azure checkaccess-v2-go-sdk
 	useCheckAccessV2     bool
 	pdpClient            checkaccess.RemotePDPClient
-	pdpScope             string
 	checkAccessV2Version string
 
 	tokenProvider                          graph.TokenProvider
@@ -233,7 +232,6 @@ func newAccessInfo(tokenProvider graph.TokenProvider, rbacURL *url.URL, opts aut
 		}
 
 		u.pdpClient = pdpClient
-		u.pdpScope = opts.PDPScope
 	}
 
 	return u, nil


authz/providers/azure/rbac/checkaccess_v2.go
```diff
@@ -0,0 +1,272 @@
...
+
+		return &authzv1.SubjectAccessReviewStatus{
+			Allowed: true,
+			Reason:  verdict,
+			Denied:  false,
+		}
+	}
+
+	// Access denied
+	if denyReason == "" {
```
chrischangcode: Question for my understanding: It looks like this would happen if `len(decisions) == 0`. Is that an expected state if we requested for a non-empty batch of actions? Would that mean no access conditions are set at all and without any, access defaults to denied?

weinong:
@arxhive can you rebase? it has changes that are already merged

authz/providers/azure/rbac/tokencredential_adapter.go
```diff
@@ -0,0 +1,60 @@
...
+	provider graph.TokenProvider
+	scope    string
+}
+
+// newTokenProviderAdapter creates a new adapter that wraps a graph.TokenProvider
+// and implements azcore.TokenCredential interface.
+func newTokenProviderAdapter(provider graph.TokenProvider, scope string) azcore.TokenCredential {
+	return &tokenProviderAdapter{
+		provider: provider,
+		scope:    scope,
```
weinong: this isn't used

authz/providers/azure/rbac/rbac.go
```diff
@@ -376,57 +421,51 @@ func (a *AccessInfo) performCheckAccess(
...
+func (a *AccessInfo) CheckAccess(ctx context.Context, request *authzv1.SubjectAccessReviewSpec) (*authzv1.SubjectAccessReviewStatus, error) {
+	a.auditSARIfNeeded(ctx, request)
 
-	checkAccessBodies, err := prepareCheckAccessRequestBody(request, a.clusterType, a.azureResourceId, a.useNamespaceResourceScopeFormat, a.allowCustomResourceTypeCheck, a.allowSubresourceTypeCheck)
+	log := klog.FromContext(ctx)
+
+	// Route to v2 API if enabled
+	if a.useCheckAccessV2 {
+		log.V(5).Info("Using CheckAccess v2 API")
+		return a.checkAccessV2(ctx, request)
```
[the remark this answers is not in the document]
arxhive: Added defensive `buildResourceIDForV2` function with invariant validation similar to `buildCheckAccessURL` from PR #415. The function validates:
1. baseResourceID must not be empty
2. Path traversal is prevented by verifying the result starts with baseResourceID

Updated `checkAccessV2` to use this function for all resource ID constructions (primary, managed namespace, fleet).

authz/providers/azure/options/options.go
```diff
@@ -54,6 +54,11 @@ type Options struct {
 	KubeConfigFile                         string
 	AuditSAR                               bool
 	FleetManagerResourceId                 string
+	// CheckAccess V2 API configuration
+	UseCheckAccessV2 bool
+	PDPEndpoint      string
```
weinong: Scope and endpoint are not the same.

authz/providers/azure/rbac/rbac.go
```diff
@@ -197,6 +206,27 @@ func newAccessInfo(tokenProvider graph.TokenProvider, rbacURL *url.URL, opts aut
...
+		// Create token credential adapter from existing token provider
+		scope := fmt.Sprintf("%s.default", authzInfo.ARMEndPoint)
+		tokenCred := newTokenProviderAdapter(tokenProvider, scope)
+
+		// Initialize PDP client
+		pdpClient, err := checkaccess.NewRemotePDPClient(
+			opts.PDPEndpoint,
+			scope,
+			tokenCred,
+			&azcore.ClientOptions{},
```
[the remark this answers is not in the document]
weinong: I believe you can pass in http client to the `azcore.ClientOptions{}` below. Possibly the headers, too. The point is whether we can retain the capability to customize the user agent. I believe it can.

authz/providers/azure/rbac/checkaccess_v2.go
```diff
@@ -0,0 +1,319 @@
...
+	if baseResourceID == "" {
+		return "", fmt.Errorf("baseResourceID must not be empty")
+	}
+
+	resourceID := baseResourceID
+	if hasNamespace {
+		resourceID = path.Join(resourceID, namespacePath)
+	}
+
+	// invariant 3: ensure no path traversal occurred
```
weinong: reading the comment... where is invariant 2? 😏

authz/providers/azure/rbac/checkaccess_v2.go
```diff
@@ -0,0 +1,319 @@
...
+		batchActions := actions[i:end]
+		correlationID := uuid.New().String()
+		batchLog := log.WithValues("correlationID", correlationID, "batchIndex", batchIndex, "actionsCount", len(batchActions))
+		batchCtx := klog.NewContext(ctx, batchLog)
+
+		batchLog.V(7).Info("Starting CheckAccess v2 batch")
+		start := time.Now()
+
+		// Create authorization request using v2 SDK helper
+		authzReq, err := a.pdpClient.CreateAuthorizationRequest(resourceId, batchActions, jwtToken)
```
weinong: This function will not work for AKS scenario as the passed-in `jwtToken` does not have the claim or `oid` to properly setup the `authzReq`.
Essentially, you need to replicate what we do [here](https://github.com/kubeguard/guard/blob/110de5d7538706cb629984df63947b8fbf68702a/authz/providers/azure/rbac/checkaccessreqhelper.go#L674C2-L684C46)

authz/providers/azure/rbac/rbac.go
```diff
@@ -197,6 +206,27 @@ func newAccessInfo(tokenProvider graph.TokenProvider, rbacURL *url.URL, opts aut
 
 	u.lock = sync.RWMutex{}
 
+	// Initialize CheckAccess V2 client if enabled
+	if opts.UseCheckAccessV2 {
+		// Create token credential adapter from existing token provider
+		scope := fmt.Sprintf("%s.default", authzInfo.ARMEndPoint)
```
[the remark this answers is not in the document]
arxhive: Fixed. Added new `--azure.pdp-scope` configuration flag that must be set explicitly. The scope should be `https://authorization.azure.net/.default` for public cloud or `https://authorization.azure.us/.default` for US Government. Removed the incorrect derivation from ARM endpoint.

```go
// In options.go
PDPScope string

// In rbac.go - now uses opts.PDPScope directly
pdpClient, err := checkaccess.NewRemotePDPClient(
    opts.PDPEndpoint,
    opts.PDPScope,  // User-provided scope
    tokenCred,
    clientOpts,
)
```

authz/providers/azure/options/options.go
arxhive: Added `--azure.pdp-scope` flag to distinguish from `--azure.pdp-endpoint`. Validation ensures both are required when v2 is enabled:

```go
// options.go
PDPEndpoint string  // PDP service URL
PDPScope    string  // OAuth scope (e.g., https://authorization.azure.net/.default)

// Validation
if o.UseCheckAccessV2 && o.PDPScope == "" {
    errs = append(errs, errors.New("azure.pdp-scope must be non-empty when azure.use-checkaccess-v2 is enabled"))
}
```

authz/providers/azure/rbac/rbac.go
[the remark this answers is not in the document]
arxhive: Fixed. Configured `azcore.ClientOptions.Transport` with existing `httpclient.DefaultHTTPClient` to retain HTTP customization capability:

```go
clientOpts := &azcore.ClientOptions{
    Transport: httpclient.DefaultHTTPClient,
}

pdpClient, err := checkaccess.NewRemotePDPClient(
    opts.PDPEndpoint,
    opts.PDPScope,
    tokenCred,
    clientOpts,
)
```

This allows the v2 SDK to use the same HTTP client configuration as v1, preserving any custom transport settings.

authz/providers/azure/rbac/checkaccess_v2.go
arxhive: Fixed. The SDK's `CreateAuthorizationRequest` extracts oid/groups from JWT claims, which does not work for AKS where user identity comes from `SubjectAccessReviewSpec.Extra["oid"]` and `SubjectAccessReviewSpec.Groups`.

Added two new functions to handle this:

```go
// extractUserIdentityV2 extracts oid and groups from request
// Mirrors v1 logic in checkaccessreqhelper.go:674-684
func extractUserIdentityV2(request *authzv1.SubjectAccessReviewSpec) (userOid string, groups []string, err error) {
    if oid, ok := request.Extra["oid"]; ok {
        userOid = val[1 : len(val)-1]  // Remove brackets
        if !isValidUUID(userOid) {
            return "", nil, errutils.WithCode(...)
        }
    } else {
        return "", nil, errutils.WithCode(errors.New("oid info not sent"))
    }
    groups = getValidSecurityGroups(request.Groups)
    return userOid, groups, nil
}

// buildAuthorizationRequestV2 constructs request manually
func buildAuthorizationRequestV2(resourceId string, actions []string, userOid string, groups []string) checkaccess.AuthorizationRequest {
    return checkaccess.AuthorizationRequest{
        Subject: checkaccess.SubjectInfo{
            Attributes: checkaccess.SubjectAttributes{
                ObjectId: userOid,
                Groups:   groups,
            },
        },
        Actions:  actionInfos,
        Resource: checkaccess.ResourceInfo{Id: resourceId},
    }
}
```

The `performCheckAccessV2` function now accepts `userOid` and `groups` directly instead of `jwtToken`.

authz/providers/azure/rbac/rbac.go
```diff
@@ -195,6 +207,35 @@ func newAccessInfo(tokenProvider graph.TokenProvider, rbacURL *url.URL, opts aut
...
+			opts.PDPScope,
+			tokenCred,
+			clientOpts,
+		)
+		if err != nil {
+			return nil, fmt.Errorf("failed to create CheckAccess v2 client: %w", err)
+		}
+
+		u.pdpClient = pdpClient
+		u.pdpScope = opts.PDPScope
```
weinong: this is no longer used, i believe

arxhive: Agree, we can cleanup `pdpScope`.

But `pdpClient` is still used - called in checkaccess_v2.go:97 for CheckAccess() calls

PR #432, opened 2025-12-31
fix: check fleet scope first when fleetResourceID exists

jwtty:
For a managedCluster/connectedCluster joining fleet, we call it a fleet member cluster, and its guard configuration has the `fleetResourceID` set. 
We have a list of "for Member Cluster" roles: https://learn.microsoft.com/en-us/azure/kubernetes-fleet/concepts-rbac#kubernetes-data-plane (e.g. "Azure Kubernetes Fleet Manager RBAC Cluster Admin for Member Clusters") that can be granted at fleet or fleet managedNamespace scope so that the assginee can have the same access across all member clusters with just a single role assignment. 

Existing code behavior is:
When the user request is cluster scoped, e.g. creating a namespace or listing pods across all namespaces or getting a cluster-scoped resource, guard checks access at fleet level: `POST https://management.azure.com/subscriptions/<REDACTED>/resourceGroups/<REDACTED>/providers/Microsoft.ContainerService/fleets/<FLEET_NAME>/providers/Microsoft.Authorization/checkaccess?api-version=2018-09-01-preview`;

on the other hand, when the user request is namespace scoped, we check against fleet managedNamespace ONLY: `POST https://management.azure.com/subscriptions/<REDACTED>/resourceGroups/<REDACTED>/providers/Microsoft.ContainerService/fleets/<FLEET_NAME>/managedNamespaces/kube-system/providers/Microsoft.Authorization/checkaccess?api-version=2018-09-01-preview`

This is where we have the issue: when granting "Azure Kubernetes Fleet Manager RBAC Cluster Admin for Member Clusters", user should be able to make namespace scoped request, e.g. listing pods in one namespace but at the moment the request is denied because the fleet managedNamespace azure resource does not exist.

This PR fixes this issue by checking against both fleet scope and fleet managedNamespace scope when the request is namespace scoped. The workflow is updated to:
1. Always first check against fleet scope.
2. If does not pass and the request is namespace scoped, check against fleet managedNamespace. 
This aligns with existing managedCluster checkAccess logic.

### DIFF
--- a/authz/providers/azure/rbac/checkaccess_v2.go
+++ b/authz/providers/azure/rbac/checkaccess_v2.go
@@ -261,8 +261,8 @@ func (a *AccessInfo) checkAccessV2(ctx context.Context, request *authzv1.Subject
 	}
 
 	// Determine resource ID (with or without namespace scope)
-	namespaceExist, namespaceString := getNameSpaceScope(request, a.useNamespaceResourceScopeFormat)
-	resourceId, err := buildResourceIDForV2(a.azureResourceId, namespaceExist, namespaceString)
+	namespaceExists, namespaceString := getNameSpaceScope(request, a.useNamespaceResourceScopeFormat)
+	resourceId, err := buildResourceIDForV2(a.azureResourceId, namespaceExists, namespaceString)
 	if err != nil {
 		return nil, fmt.Errorf("error building primary resource ID: %w", err)
 	}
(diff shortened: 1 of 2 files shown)


authz/providers/azure/rbac/checkaccess_v2.go
```diff
@@ -302,7 +302,7 @@ func (a *AccessInfo) checkAccessV2(ctx context.Context, request *authzv1.Subject
 	if a.fleetManagerResourceId != "" {
 		log.V(7).Info("Falling back to fleet manager scope check (v2)", "fleetResourceId", a.fleetManagerResourceId)
 
-		fleetResourceId, err := buildResourceIDForV2(a.fleetManagerResourceId, managedNamespaceExists, managedNamespacePath)
+		fleetResourceId, err := buildResourceIDForV2(a.fleetManagerResourceId, namespaceExist, namespaceString)
```
weng271190436: The PR description says

"When fleetResourceID exists, always check against the fleet scope first. Then if the request is namespace scoped, check against fleet managedNamespace secondly."

by "fleet scope," do you mean "/fleets/test-fleet/namespaces/test-ns" or "/fleets/test-fleet"

The description seems to imply the latter but code seems to imply the former

jwtty: It depends on the request, if the request is namespace scoped, it checks `/fleet/test-fleet/namespace/test-ns`; otherwise, it checks `/fleet/test-fleet`.

weinong (approved):
LGTM

PR #433, opened 2026-01-05
fix: use fleetMembers actions for v2 fleet scope authorization checks

arxhive:
## Summary

This PR fixes the remaining issue by ensuring that when falling back to fleet scope checks, the v2 implementation generates fleet-specific actions using the `fleetMembers` cluster type, consistent with how v1 handles this scenario.

### The Problem

In the v2 `checkAccessV2` function, actions were computed once at the beginning using `a.clusterType`:
```go
actions, err := getDataActionsV2(ctx, request, a.clusterType, ...)
```

These same `actions` were then reused for fleet scope checks, which is incorrect. For a fleet member cluster with `clusterType = managedClusters`, this would send:
- `Microsoft.ContainerService/managedClusters/pods/read`

Instead of the correct fleet member action:
- `Microsoft.ContainerService/fleets/members/pods/read`

### The Fix

Generate fleet-specific actions using `fleetMembers` cluster type before performing fleet scope checks:
```go
fleetActions, err := getDataActionsV2(ctx, request, fleetMembers, ...)
```

This ensures roles like "Azure Kubernetes Fleet Manager RBAC Cluster Admin for Member Clusters" can properly authorize requests through the v2 API.

### Changes

- Generate `fleetMembers` actions for fleet scope and fleet managed namespace checks in v2
- Add test `TestCheckAccessV2_FleetActionsUseCorrectClusterType` to verify correct actions are used

### DIFF
--- a/authz/providers/azure/rbac/checkaccess_v2.go
+++ b/authz/providers/azure/rbac/checkaccess_v2.go
@@ -305,7 +305,7 @@ func (a *AccessInfo) checkAccessV2(ctx context.Context, request *authzv1.Subject
 		// Generate fleet-specific actions using fleetMembers cluster type
 		// This ensures actions like "Microsoft.ContainerService/fleets/members/pods/read"
 		// are used instead of "Microsoft.ContainerService/managedClusters/pods/read"
-		fleetActions, err := getDataActionsV2(ctx, request, fleetMembers, a.allowCustomResourceTypeCheck, a.allowSubresourceTypeCheck)
+		fleetMemberActions, err := getDataActionsV2(ctx, request, fleetMembers, a.allowCustomResourceTypeCheck, a.allowSubresourceTypeCheck)
 		if err != nil {
 			return nil, fmt.Errorf("error preparing fleet actions list: %w", err)
 		}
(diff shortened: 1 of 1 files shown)


authz/providers/azure/rbac/checkaccess_v2.go
```diff
@@ -302,13 +302,20 @@ func (a *AccessInfo) checkAccessV2(ctx context.Context, request *authzv1.Subject
 	if a.fleetManagerResourceId != "" {
 		log.V(7).Info("Falling back to fleet manager scope check (v2)", "fleetResourceId", a.fleetManagerResourceId)
 
+		// Generate fleet-specific actions using fleetMembers cluster type
+		// This ensures actions like "Microsoft.ContainerService/fleets/members/pods/read"
+		// are used instead of "Microsoft.ContainerService/managedClusters/pods/read"
+		fleetActions, err := getDataActionsV2(ctx, request, fleetMembers, a.allowCustomResourceTypeCheck, a.allowSubresourceTypeCheck)
```
jwtty:
```suggestion
		fleetMemberActions, err := getDataActionsV2(ctx, request, fleetMembers, a.allowCustomResourceTypeCheck, a.allowSubresourceTypeCheck)
```

PR #445, opened 2026-04-08
## Short-circuit service principal group overage with clear error message

chrischangcode:
### Problem

When a service principal (SPN) with >200 group memberships authenticates to an AKS cluster, Azure AD returns a group overage claim instead of embedding groups in the token. Guard then attempts to resolve group membership via the Graph API using the On-Behalf-Of (OBO) flow - but OBO is not supported for service principal tokens, causing Azure AD to return `AADSTS7000113`. The resulting error message was cryptic and gave no actionable guidance.

### Root Cause

Guard did not detect that a token belonged to a service principal before attempting the OBO-based Graph API call. When the OBO exchange failed, the raw `AADSTS7000113` error surfaced with no context about why it happened or how to resolve it.

#### 1. Short-circuit SPN group overage (auth path)

- **`auth/providers/azure/azure.go`**: Added `isAppToken(claims)` helper that checks the `idtyp` claim (`"app"` = service principal, absent = v1 user token). When overage is detected and the token is from a SPN, `Check()` returns immediately with a clear error message and link to the [kubelogin limitations documentation](https://learn.microsoft.com/en-us/azure/aks/kubelogin-authentication#kubelogin-authentication-in-aks-limitations). The error is wrapped with `errutils.WithCode(http.StatusOK)` so that Guard returns HTTP 200, which causes the API server's webhook authenticator to read and log the `Status.Error` message rather than treating the response as a transport failure.

- **`auth/providers/azure/graph/graph.go`**: Extended existing ARC path SPN check with `errutils.WithCode(http.StatusOK)` to ensure consistent error visibility.

#### 2. Tests

- **`auth/providers/azure/azure_test.go`**: `TestCheckAzureAuthenticationSPNWithOverage` (2 sub-tests: SPN+overage blocked, SPN+groups-in-token succeeds), `TestIsAppToken` (3 sub-tests).

### Error visibility

| Consumer | Before | After |
|----------|--------|-------|
| Guard logs | Cryptic `AADSTS7000113` from failed OBO exchange | Clear message: *"service principal with group membership exceeding 200 is not supported"* with docs link |
| API server logs | No — HTTP 401 caused webhook authenticator to skip `Status.Error` body | Yes — HTTP 200 lets webhook authenticator read and log `Status.Error` via `klog.ErrorS(err, "Unable to authenticate the request")` |
| End user (kubectl) | Hardcoded `"Unauthorized"` | Hardcoded `"Unauthorized"` (unchanged — Kubernetes security design) |

### Files Changed (4 files, +266/-3)

- `auth/providers/azure/azure.go` — `isAppToken()`, early-exit in `Check()` with `WithCode(200)`
- `auth/providers/azure/azure_test.go` — SPN overage + isAppToken tests
- `auth/providers/azure/graph/graph.go` — ARC path SPN check wrapped with `WithCode(200)`
- `auth/providers/azure/graph/graph_test.go` — assertion message update


auth/providers/azure/graph/graph.go
```diff
@@ -218,7 +219,10 @@ func (u *UserInfo) getMemberGroupsUsingARCOboService(ctx context.Context, access
 		}
 	}
 	if !isAADUser {
-		return nil, errors.New("Overage claim (users with more than 200 group membership) for SPN is currently not supported. For troubleshooting, please refer to aka.ms/overageclaimtroubleshoot")
+		return nil, errutils.WithCode(
+			fmt.Errorf("Overage claim (users with more than 200 group membership) for SPN is currently not supported. "+
+				"For troubleshooting, please refer to aka.ms/overageclaimtroubleshoot"),
+			http.StatusOK)
```
weinong, line 233: it'd be good to add a comment about why we return error with status OK

auth/providers/azure/azure_test.go
```diff
@@ -412,6 +415,66 @@ func TestCheckAzureAuthenticationWithOverageCheckOption(t *testing.T) {
...
+		defer srv.Close()
+
+		token, err := signKey.sign([]byte(fmt.Sprintf(accessTokenSPNWithOverage, srv.URL)))
+		if err != nil {
+			t.Fatalf("Error when signing token. reason: %v", err)
+		}
+
+		resp, err := client.Check(ctx, token)
+		assert.NotNil(t, err)
+		assert.Nil(t, resp)
```
weinong, line 438: can we add the assertion about the status OK?

weinong (changes requested):
mostly lgtm with minor feedbacks

auth/providers/azure/graph/graph.go
chrischangcode: that's a good point! added for both `graph.go` and `azure.go`

auth/providers/azure/azure_test.go
chrischangcode: Thanks for the feedback I added the assertion for both test files!

weinong (approved):
LGTM

PR #447, opened 2026-04-21
Support using the Entra SDK sidecar container to validate Azure access tokens

### LINKED ISSUE #448: Tracking issue: Use Entra SDK sidecar container to validate Azure access tokens
There are some systems using Guard that want to perform the Azure access token validation using the "official" dotnet implementation. This is where the Entra SDK container comes in:
https://learn.microsoft.com/en-us/entra/msidweb/agent-id-sdk/overview

https://mcr.microsoft.com/en-us/artifact/mar/entra-sdk/auth-sidecar/about

Guard should add support for using this container as a sidecar to validate Azure access tokens. The container cannot fully validate PoP tokens, so it can be used just for the inner access token validation in that scenario.

AzureMarker:
NOTE: Most of the changed lines are due to vendoring some packages for the new E2E tests.

There are some systems using Guard that want to perform the Azure access token validation using the "official" dotnet implementation. This is where the Entra SDK container comes in:
https://learn.microsoft.com/en-us/entra/msidweb/agent-id-sdk/overview

https://mcr.microsoft.com/en-us/artifact/mar/entra-sdk/auth-sidecar/about

This PR adds support for using this container as a sidecar for Azure access token validation. It cannot fully validate PoP tokens, so it's only used for the inner access token validation in that scenario.

This feature is enabled only when the new `--azure.entra-sdk-url` flag is set in the Guard options, for example:
```
--azure.entra-sdk-url=http://localhost:8080
```

The PR also fixes up the E2E tests and adds some that run through the full Azure token validation flow. This required some changes in the installer as well.

I added a local E2E testing script that automates the work of building the container image locally, loading it into the kind cluster, and running the E2E with that container image. There is a readme at `test/e2e/README.md` that explains how to use the script.

Closes #448


installer/deployment.go
```diff
@@ -77,7 +88,7 @@ func newDeployment(authopts AuthOptions, authzopts AuthzOptions) (objects []runt
 										Scheme: core.URISchemeHTTPS,
 									},
 								},
-								InitialDelaySeconds: int32(30),
+								InitialDelaySeconds: 1,
```
weinong, line 86: this change will be a behavior change. though, it's unclear who's really using the installer. but is it really necessary?

AzureMarker: It's not necessary, but a 30 second wait is unnecessary and slows down the tests a lot.

weinong:
generally LGTM. I pinged few people from AKS to review as well. We can aim to merge it later this week if they don't have other feedback.

auth/providers/azure/verifier.go
```diff
@@ -0,0 +1,288 @@
...
+
+var (
+	_ AccessTokenVerifier = (*OIDCAccessTokenVerifier)(nil)
+	_ AccessTokenVerifier = (*EntraSDKTokenVerifier)(nil)
+	_ VerifiedAccessToken = (*oidcVerifiedAccessToken)(nil)
+	_ VerifiedAccessToken = (*staticClaimsToken)(nil)
+)
+
+type OIDCAccessTokenVerifier struct {
+	Verifier *oidc.IDTokenVerifier
```
norshtein: nit: do not expect this `Verifier` be used outside of the package, should not be exported?

norshtein:
Seems the CI does not contain E2E, do we have an E2E test result for the new code?

installer/options.go
```diff
@@ -121,7 +128,22 @@ func (o *AuthOptions) Validate() []error {
...
+		errs = append(errs, azureOpts.Validate()...)
+		if o.UseAzureEntraSDK {
+			if _, err := o.Azure.EntraSDKEnvVars(); err != nil {
+				errs = append(errs, err)
+			}
+		}
+	} else if o.UseAzureEntraSDK {
+		errs = append(errs, errors.New("azure.use-entra-sdk requires azure auth provider"))
+	}
+	if o.UseAzureEntraSDK && o.AzureEntraSDKImage == "" {
```
norshtein, line 145: I assume this line will always evaluate to false since `o.AzureEntraSDKImage` has default value?

installer/options.go
```diff
@@ -121,7 +128,22 @@ func (o *AuthOptions) Validate() []error {
...
 	}
 	if o.AuthProvider.Has(azure.OrgType) {
-		errs = append(errs, o.Azure.Validate()...)
+		azureOpts := o.Azure
+		if o.UseAzureEntraSDK {
+			// The SDK URL will be generated
+			azureOpts.EntraSDKURL = ""
+		}
+		errs = append(errs, azureOpts.Validate()...)
+		if o.UseAzureEntraSDK {
```
norshtein, line 137: nit: duplicate `if` statement in L132, can merge them into 1. The `if` statement in L142 can also be merged.

norshtein:
My comments are mainly about nit and those are non-blocking. The code's functionality LGTM.

installer/options.go
```diff
@@ -90,7 +94,10 @@ func (o *AuthOptions) AddFlags(fs *pflag.FlagSet) {
 	fs.StringVarP(&o.Namespace, "namespace", "n", o.Namespace, "Name of Kubernetes namespace used to run guard server.")
 	fs.StringVar(&o.Addr, "addr", o.Addr, "Address (host:port) of guard server.")
 	fs.BoolVar(&o.RunOnMaster, "run-on-master", o.RunOnMaster, "If true, runs Guard server on master instances")
+	fs.StringVar(&o.GuardImage, "guard-image", o.GuardImage, "Container image for the Guard server. When set, this overrides --private-registry and the derived version tag")
 	fs.StringVar(&o.PrivateRegistry, "private-registry", o.PrivateRegistry, "Private Docker registry")
+	fs.BoolVar(&o.UseAzureEntraSDK, "azure.use-entra-sdk", o.UseAzureEntraSDK, "If true, configures Guard installer to run the Entra SDK sidecar for Azure auth")
```
karataliu, line 99: merge into o.Azure.AddFlags(fs) below since azure specific?

installer/deployment.go
```diff
@@ -38,6 +39,16 @@ import (
 	"k8s.io/apimachinery/pkg/util/intstr"
 )
 
+const (
```
karataliu, line 42: this is top level shared by all providers, consider a separate file for azure specific things.

auth/providers/azure/verifier.go
```diff
@@ -0,0 +1,288 @@
...
+}
+
+type EntraSDKTokenVerifier struct {
+	baseURL              *url.URL
+	clientID             string
+	verifyClientID       bool
+	httpClientRetryCount int
+}
+
+func newEntraSDKTokenVerifier(rawBaseURL, clientID string, verifyClientID bool, httpClientRetryCount int) (*EntraSDKTokenVerifier, error) {
```
karataliu: this is detailed entrasdk impl, move to entrasdk_verifier.go?
this file suits common interface

installer/options.go
AzureMarker: The user can override the default via `--azure.entra-sdk-image=""`

installer/options.go
AzureMarker: The if statements can't be merged because we need to run `azureOpts.Validate` in-between the two operations.
The L142 if statement also can't be merged because it handles the case when the auth provider is not Azure.

installer/options.go
AzureMarker: This is an installer-specific flag, so it cannot be moved into the main auth provider flags.

AzureMarker:
> Seems the CI does not contain E2E, do we have an E2E test result for the new code?

@norshtein Yes, the E2E was successful when I ran it locally:
image

There's an old PR to add the E2Es to PR checks: https://github.com/kubeguard/guard/pull/232

auth/providers/azure/verifier.go
AzureMarker: Updated

installer/deployment.go
AzureMarker: Updated

auth/providers/azure/verifier.go
AzureMarker: Updated

norshtein (approved):

installer/deployment.go
```diff
@@ -118,16 +132,58 @@ func newDeployment(authopts AuthOptions, authzopts AuthzOptions) (objects []runt
...
+	}
 	if authopts.HttpsProxy != "" || authopts.HttpProxy != "" || authopts.NoProxy != "" {
 		proxyEnvVarRef := core.EnvFromSource{
 			SecretRef: &core.SecretEnvSource{},
 		}
 
 		proxyEnvVarRef.SecretRef.Name = "guard-proxy"
 		d.Spec.Template.Spec.Containers[0].EnvFrom = append(d.Spec.Template.Spec.Containers[0].EnvFrom, proxyEnvVarRef)
+		if authopts.UseAzureEntraSDK {
+			d.Spec.Template.Spec.Containers[1].EnvFrom = append(d.Spec.Template.Spec.Containers[1].EnvFrom, proxyEnvVarRef)
```
arxhive, line 177: `Containers[1]` is accessed here under the condition `authopts.UseAzureEntraSDK`, but the Entra SDK container is only appended at line 140 when `authopts.AuthProvider.Has(azure.OrgType) && authopts.UseAzureEntraSDK`. If `UseAzureEntraSDK=true` but the azure auth provider is absent (with proxy configured), this panics with index-out-of-bounds.

While `Validate()` catches this misconfiguration upstream, `newDeployment` itself is not defensive - callers can reach `Generate()` without calling `Validate()` first.

The same applies to line 252 where `Containers[1].VolumeMounts` is accessed under the same weaker condition.

**Suggestion**: Match the condition used at line 135, or look up the container by name:

```suggestion
		if authopts.AuthProvider.Has(azure.OrgType) && authopts.UseAzureEntraSDK {
			d.Spec.Template.Spec.Containers[1].EnvFrom = append(d.Spec.Template.Spec.Containers[1].EnvFrom, proxyEnvVarRef)
		}
```

auth/providers/azure/verifier.go
```diff
@@ -0,0 +1,29 @@
...
+limitations under the License.
+*/
+
+package azure
+
+import (
+	"context"
+)
+
+type AccessTokenVerifier interface {
```
arxhive, line 23: `MakeRetryableHttpClient` is called on every `Verify()` invocation (line 133 of this file). Since `Verify` runs on every authentication request (the webhook hot path for every kubectl action), this allocates a new `retryablehttp.Client` struct per call unnecessarily.

While the underlying `http.Transport` is shared via the `DefaultHTTPClient` copy, the `retryablehttp.Client` and its `contextAwareLogger` (which captures `ctx`) are heap-allocated each time.

**Suggestion**: Build the client once in `newEntraSDKTokenVerifier` and store it as a field on `EntraSDKTokenVerifier`. The per-request context is already threaded via `req.WithContext(ctx)`.

auth/providers/azure/verifier.go
```diff
@@ -0,0 +1,29 @@
...
+	Verify(ctx context.Context, rawAccessToken string) (VerifiedAccessToken, error)
+}
+
+type VerifiedAccessToken interface {
+	Claims() (claims, error)
```
arxhive, line 28: `entraSDKValidateResponse.Protocol` and `entraSDKValidateResponse.Token` (lines 212-213) are deserialized from JSON but never accessed anywhere in the code - only `Claims` is used. Similarly, `entraSDKErrorResponse.Status` (line 221) is parsed but the code uses the HTTP response `statusCode` parameter for all status-based decisions.

These dead fields add noise to the API contract. Consider removing them:

```go
type entraSDKValidateResponse struct {
    Claims map[string]interface{} `json:"claims"`
}

type entraSDKErrorResponse struct {
    Type   string `json:"type"`
    Title  string `json:"title"`
    Detail string `json:"detail"`
}
```

auth/providers/azure/verifier.go
AzureMarker: `EntraSDKTokenVerifier` and the rest of the `azure.Authenticator` data is built once for every request, so there is no benefit to moving the `MakeRetryableHttpClient` call. The work and allocations it performs are minimal and acceptable.

auth/providers/azure/verifier.go
AzureMarker: Updated

installer/deployment.go
AzureMarker: Updated

arxhive:
Thank you for the updates, LGTM!

arxhive (approved):

PR #450, opened 2026-05-09
fix: tolerate degraded aggregated APIServices during API resource discovery

arxhive:
## Summary

- Guard fatally crashes on startup when aggregated APIServices return 0-length HTTP 200 responses during API resource discovery, causing a crash loop that prevents Guard readiness
- Broadened the error handling in `fetchApiResources()` to tolerate all partial API discovery errors (not just "unreachable" errors), since `ServerPreferredResources()` always returns valid partial results alongside these errors
- Removed the now-unused `apiserviceUnreachableError` constant

## Context

**ICM 792913310** - 59 clusters in centralus have Guard crash-looping because `nodeagent.microsoft.com/v1alpha1` aggregated APIService returns 0-length HTTP 200 responses. Guard (`server.go:261`) fatally exits when API resource discovery fails, causing 6,365 crashes in 6 hours. This cascades into OverlayManager `UpgradeRelease` timeouts since Guard readiness is a precondition for Helm release upgrades.

The existing error handling checked for `retrieveServerApiError` AND `apiserviceUnreachableError`, but the 0-length response error only matched the first condition. The `ServerPreferredResources()` function returns partial results whenever the `retrieveServerApiError` prefix is present, so checking for that prefix alone is sufficient and more resilient to future error patterns.

**Affected API groups in production:**
- `nodeagent.microsoft.com/v1alpha1` (3,677 crashes) - Azure CNI Node Agent
- `extensions.containerapp.microsoft.com/v1alpha1` (202 crashes)
- `containerapp.lgnvk.microsoft.com/v1alph`


weinong (approved):
lgtm

PR #454, opened 2026-06-02
fix: Check subresource for CSR/nodeclient

weng271190436:
When we check certificatesigningrequests data actions, we only check resource and verb but not subresource.

If someone has CRS/write, they can get access to CSR/nodeclient/write, which allows a user to impersonate as a node.


weng271190436:
closing and opening one in Azure fork https://github.com/Azure/guard/pull/12

